PR #5, opened 2019-07-01
Turn litex_boards.partner into module

### LINKED ISSUE #6: Partner platforms/target are not modules
`litex_boards.partner`, `litex_boards.partner.platforms` and `litex_boards.partner.targets` are not valid Python modules as they lack `__init__.py`.

DurandA:
Fix #6


enjoy-digital:
Thanks.

PR #13, opened 2019-08-22
Fix ecp5_evn clock

DurandA:
The system clock was driven directly while it should be driven by the PLL. I somehow forgot to remove this line before the initial PR.


enjoy-digital:
Thanks.

PR #17, opened 2019-09-02
partner: add fomu target

xobs:
This adds the Fomu target back in.  The default BaseSoC supports
various USB methods, and will be updated as more become available.

The debug bridge may optionally be added.

Signed-off-by: Sean Cross <sean@xobs.io>

### DIFF
--- a/litex_boards/partner/targets/fomu.py
+++ b/litex_boards/partner/targets/fomu.py
@@ -1,0 +1,246 @@
+#!/usr/bin/env python3
+
+# This file is Copyright (c) 2018-2019 Florent Kermarrec <florent@enjoy-digital.fr>
+# This file is Copyright (c) 2018 David Shah <dave@ds0.me>
+# License: BSD
+
+import argparse
+
+from migen import *
+from migen.genlib.resetsync import AsyncResetSynchronizer
+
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, merged.  I've fixed the copyright (since i haven't contributed anything for now :)). I'll probably do some test with the full Fomu design soon and the iCE40PLL module i added to LiteX recently: https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/clock.py#L390-L483 We could probably simplify Fomu clocking using it.

xobs:
Wow, that's an amazing module.

One thing to note about the ICE40PLL -- the documentation is apparently wrong.  They say the _minimum_ output frequency is 16 MHz, however that ignores the  `p_PLLOUT_SELECT = "GENCLK_HALF"` parameter that we use.  The minimum _full_ frequency is 16 MHz, but if you divide it by two you can run it down to 8 MHz.  Which is how we get 12 MHz out of it.

enjoy-digital:
Thanks, indeed, that's what i want to support with the following TODO: https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/clock.py#L394

PR #18, opened 2019-09-17
Fomu cpu updates

xobs:
This makes some minor changes to the Fomu target, allowing for the use of `fomu` with a CPU and an interrupt source on the USB module.


xobs:
I also added i2c pins to the definition for EVT.

PR #19, opened 2019-10-13
add sys clock freq flag, uses same method as current versa code

lolsborn:
Allows setting system clock frequency via command-line argument.  Uses the same code/method as the versa-ecp5 target.  This is useful when building rocket synthesis doesn't pass the default timing of 50mhz.


enjoy-digital:
Thanks!

PR #20, opened 2019-10-13
memory device selection for ulx3s

lolsborn:
I have 4 of these boards now.  3 that I've bought from Radiona and one that I made myself and between them they have 3 different SDRAM parts.  It's only a two-line change currently to switch between parts, but having an argument switch is a nice convenience.

If your board had something else on-board all you would have to do is extend the import statement to include your part number.

If pass an invalid / unsupported part number you get a message like: ```AttributeError: module '__main__' has no attribute 'AS4C16M16F'```

### DIFF
--- a/litex_boards/partner/targets/ulx3s.py
+++ b/litex_boards/partner/targets/ulx3s.py
@@ -5,6 +5,7 @@
 # License: BSD
 
 import argparse
+import sys
 
 from migen import *
 from migen.genlib.resetsync import AsyncResetSynchronizer
@@ -15,7 +16,7 @@ from litex.soc.cores.clock import *
 from litex.soc.integration.soc_sdram import *
 from litex.soc.integration.builder import *
 
-from litedram.modules import MT48LC16M16
+from litedram.modules import MT48LC16M16, AS4C32M16, AS4C16M16
 from litedram.phy import GENSDRPHY
 
 # CRG ----------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


litex_boards/partner/targets/ulx3s.py
```diff
@@ -62,8 +65,13 @@ def __init__(self, device="LFE5U-45F", toolchain="diamond", sys_clk_freq=int(50e
         self.submodules.crg = _CRG(platform, sys_clk_freq)
 
         if not self.integrated_main_ram_size:
-            self.submodules.sdrphy = GENSDRPHY(platform.request("sdram"), cl=2)
-            sdram_module = MT48LC16M16(sys_clk_freq, "1:1")
+            if mem_device.strip().upper() == "MT48LC16M16":
+                self.submodules.sdrphy = GENSDRPHY(platform.request("sdram"), cl=2)
```
lolsborn: This is the only thing that feels a bit dirty. If there are a lot of potential variations this could grow to be hairy.

enjoy-digital:
Thanks, that's merged with some modifications (https://github.com/litex-hub/litex-boards/commit/91083f99a8551c3465aaf3d6130268c7f7f24a50)

lolsborn:
Thanks, those changes make a lot of sense.  I'm not sure why I thought I needed that if statement since the default is cl=2 and those are functionally equivalent.

PR #32, opened 2020-01-07
Update ecp5_evn.py

DurandA:
Document that R22 and R23 should be removed to use UART.


enjoy-digital:
Thanks for the update!

PR #45, opened 2020-02-23
platforms/colorlight_5a_75b.py: fix sdram_clock and sdram a pins

trabucayre:
Fix two SDRAM pins:
- SDRAM clock is C6 not B9 (used for DQ7)
- typo in chubby75 repo take A4 connected to C1 instead of C12.

Tested with memtest


enjoy-digital:
Thanks!

PR #46, opened 2020-02-24
add new board files for VC707 and VCU118

fei-g:
Only specified limited ports for VCU118, including clock, reset and DDR4
Called uspddrphy, which is basically usddrphy with SIM_DEVICE=ULTRASCALE_PLUS


mithro:
FYI -- @kgugala

enjoy-digital:
Thanks @fei-g: this has been merged with some changes:
- https://github.com/litex-hub/litex-boards/commit/88a1f80db1cdcc39ae9843f92b41c89a960aa087
- https://github.com/litex-hub/litex-boards/commit/3581df5af677888a4f6c2dd0ec0dda2c44c08e3a
- https://github.com/litex-hub/litex-boards/commit/f279fe9d339631d8a4de7b8fea336844d94cae5b

Ultrascale Plus support has also been added to LiteDRAM with https://github.com/enjoy-digital/litedram/commit/752e006cf9b606f16152627a8c2f0b2614b752aa, using the same PHY but changing `sim_device` parameter to `"ULTRASCALE_PLUS"`.

While doing this work, i saw that your were defining: `self.add_constant("USPDDRPHY", None)` but i'm not sure you adapted the BIOS for that (it's doing some specific tasks at the initialization for Ultrascale PHYs that were probably skipped). So since LiteDRAM is using the same PHY for Ultrascale and Ultrascale Plus, i reverted `self.add_constant("USDDRPHY", None)`, which should improve things on your side.

Can you do a test with the upstream code and eventually create an issue in `litex-boards` if it's not working? @mithro and @kgugala are also interested by this work, so this would allow them to follow the progress.

PR #51, opened 2020-03-05
Add iCEBreaker FPGA support

esden:
This patch set adds the iCE40UP5k iCEBreaker FPGA development board support.

The target script currently has the bios supported as we need to remap the ROM position in the linker script so we can place the bios in the SPI FLASH. But it definitely works and we can write our own programs to run on the resulting SOC.

You can find an exmaple blinky application here: https://github.com/icebreaker-fpga/icebreaker-litex-examples


enjoy-digital:
Thanks for the board support! We can merge it now, i'll test it in the next days with your example design and see if we can get the BIOS back.

PR #56, opened 2020-03-18
de10nano add in support for MiSTer secondary sd card

rob-ng15:
The de10nano in a MiSTer setup has a secondary sd card attached to gpio pins. This patch, along with proposed patches to the litex bios allow the loading of the Linux images from this secondary sd card, partition 1 formatted as a FAT16 partition.


rob-ng15:
Support code has been merged into the bios code for Litex https://github.com/enjoy-digital/litex/pull/429

PR #58, opened 2020-03-20
Move trellisboard target to SoCCore, add SPI-mode SDCard support

gsomlo:
@enjoy-digital: this, unlike my earlier LiteSDCard based PR **works** -- successfully booted a boot.bin (bbl blob containing kernel and busybox) on the trellisboard. Please review and pull!

Thanks,
--G


gsomlo:
Updated to use LiteXSoC's new `add_spi_sdcard()` method.

PR #59, opened 2020-03-22
OrangeCrab Board updates

gregdavill:
Just some updates to the OrangeCrab board.

I've finally got around to updating the platform to support r0.2. 
I've also added options to support different device types, and different SDRAM variants.

I've got to push a module for the `MT41K512M16` to LiteDRAM soon which one of my prototype boards here has fitted.

I've removed the standard `serial` mapping, and added it as an expansion on the `GPIO` connector. I think this makes it a bit clearer that it can easily be re-mapped.

Let me know if there is anything you'd want to change?


enjoy-digital:
Thanks for the update! For the board revisions, is it ok for you if we manage the revisions in a similar way we are doing for the colorlight?: https://github.com/litex-hub/litex-boards/blob/master/litex_boards/platforms/colorlight_5a_75b.py That's just to avoid having too much platform files in the repository.  (we should probably do the same for Fomu BTW).

gregdavill:
Can do! I was using the fomu as example, I didn't realise that colorlight handled multiple versions like this.

enjoy-digital:
Thanks! Merged.

Issue #49, opened 2020-02-27
ZCU104: get ddram_64 working


enjoy-digital:
Following the discussion in https://github.com/litex-hub/litex-boards/pull/47.

Currently, only ddram_32 is working. The read leveling is not working correctly on the first 4 modules when using ddram_64.

daveshah1:
May not help much, but I know it worked around July, probably LiteX revision f4770219 or so

enjoy-digital:
@daveshah1: thanks, we'll get back to that is we don't get it working easily.
@piotr-binkowski: can you do a test with ddram_64 and provide the Write/Read leveling logs? The automatic write leveling needs to be improved and still requires some manual adjustments on some configs.

enjoy-digital:
https://github.com/litex-hub/litex-boards/issues/54#issuecomment-596436281 could also be useful to debug the ZCU104 with 64-bit.

enjoy-digital:
@piotr-binkowski: https://github.com/litex-hub/litex-boards/commit/75286f8a9b9a92e4775d0d12e3df5ed704133d58 adds a missing INTERNAL VREF on Bank 64:

![image](https://user-images.githubusercontent.com/1450143/76319699-df314a80-62df-11ea-9c04-0a7d72bc2aa5.png)

This probably explains why only half of the modules were working. Can you do a test with this?

If this is not working, can you do another test with `cmd_latency` set to 0?

piotr-binkowski:
@enjoy-digital I did the tests but it looks like LiteDRAM still has problems with read leveling
With `cmd_latency` set to 1 I get:
```
--=============== SoC ==================--
CPU:       VexRiscv @ 125MHz
ROM:       32KB
SRAM:      4KB
L2:        8KB
MAIN-RAM:  1048576KB

--========== Initialization ============--
Initializing SDRAM...
SDRAM now under software control
Write leveling:
m0: |1110000000000000000001| delay: 00
m1: |1110000000000000000001| delay: 00
m2: |1111000000000000000000| delay: 00
m3: |1111100000000000000000| delay: 00
m4: |1111111100000000000000| delay: 00
m5: |1111111000000000000000| delay: 00
m6: |1111111111000000000000| delay: 00
m7: |1111111000000000000000| delay: 00
Read leveling:
m0, b0: |00000000000000000000000000000000| delays: -
m0, b1: |00000000000000000000000000000000| delays: -
m0, b2: |11000000000000000000000000000000| delays: 13+-13
[...]

PR #60, opened 2020-03-26
zcu104: add fully working SO-DIMM config

piotr-binkowski:
Depends on enjoy-digital/litedram#172

This PR adds fully working 64-bit SO-DIMM config for ZCU104 that was tested in #49


enjoy-digital:
Great thanks.

PR #64, opened 2020-04-09
Improve compatability with colorlight 5a-75b v6.1

david-sawatzke:
Things done:
- Add `ba` pin for v6.1, since it won't build otherwise. According to the datasheet, A11 is the bank activate pin. Memtest is ok.
- Add eth_clocks 1. Simply used the pins documented in the chubby75 repo.

Also added serial pins, similar to v7.0. At 115200 baud, there was some strange behavior on the rx pin & it's loopback output, which I suspect may have to do with the capacitor that seems to be connected to the button. I changed the baud rate for the example to 9600 baud, since that works (for me).

### DIFF
--- a/litex_boards/platforms/colorlight_5a_75b.py
+++ b/litex_boards/platforms/colorlight_5a_75b.py
@@ -19,6 +19,15 @@ _io_v6_1 = [ # Documented by @smunaut
     # btn
     ("user_btn_n", 0, Pins("R16"), IOStandard("LVCMOS33")),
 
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @david-sawatzke for the PR, it's merged. I just did minor changes in the comments and added revision selection: https://github.com/litex-hub/litex-boards/commit/467b14a0ad0e9d6f84f52288c2b24b913b88a165.

PR #67, opened 2020-04-14
Addition of USB ACM for ECP5

mubes:
Additions for USB support based on Gregs' valentyusb.

### DIFF
--- a/litex_boards/platforms/colorlight_5a_75b.py
+++ b/litex_boards/platforms/colorlight_5a_75b.py
@@ -172,6 +172,13 @@ _io_v7_0 = [ # Documented by @miek
         Subsignal("tx_data", Pins("T14 R12 R13 R14")),
         IOStandard("LVCMOS33")
     ),
+
+    ("usb", 0,
+        Subsignal("d_p", Pins("M8")),
+        Subsignal("d_n", Pins("R2")),
+        Subsignal("pullup", Pins("P4")),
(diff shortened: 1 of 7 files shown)


enjoy-digital:
Thanks, it's merged with some changes: https://github.com/litex-hub/litex-boards/commit/cb959628506e61cae2d7be05276bba8f42138cfb. We have to think the best way to support others programmer and flashing, so i removed it for now. I would like to avoid specific README for each target so moved the documentation to the target itself as we are doing for the others targets. 

Since we have to keep things almost similar between targets in LiteX-Boards i hope you understand these changes. If you think it's not describing enough the board support or additional features, we could include a link in the target to a more complete example maintained externally as we are doing for the iCEBreaker for example.

mubes:
No problem with these changes, they make perfect sense. If I'd thought 
about it I would have put the comments into the files rather than 
separate readmes anyway!

On the subject of flashing - It is actually quite difficult to find good 
information on how to flash so it is important to point to a documented 
solution for that, ideally sooner rather than later.  colourlight is now 
inconsistent though because it still has _some_ load functionality in 
it. It would probably be better to remove that as it only confuses 
matters, then provide a more 'standard' loader somewhere.

Regards

DAVE

enjoy-digital:
Thanks for the feedback, i'm indeed aware of the current inconsistency on colorlight regarding loading and will try to work soon on a better support to load/flash boards.
Florent

PR #74, opened 2020-05-07
targets/orangecrab.py: propagate command arguments

### LINKED ISSUE #72: `./orangecrab.py --revision r0.1 --device 85F` generates code for a 25F device
Following the directions from https://github.com/enjoy-digital/litex/blob/master/README.md and executing the target `./orangecrab.py --revision r0.1 --device 85F` resulted in a
`./soc_basesoc_orangecrab/gateware/build_top.sh` with
```
# Autogenerated by LiteX / git: 34f26868
set -e
yosys -l top.rpt top.ys
nextpnr-ecp5 --json top.json --lpf top.lpf --textcfg top.config      --25k --package CSFBGA285 --speed 8 --timing-allow-fail 
ecppack top.config --svf top.svf --bit top.bit
```

The `--25k` should be `--85k`

When fixing this manually the result builds and runs, but SDRAM test fails, so I suspect the revision also isn't propagated correctly.

tommythorn:
The parsed args are stripped off by soc_core_argdict() (called from
soc_sdram_argdict() so we have to pass them explicitly (or pass the
original "args", but this mimics the rest of the code in the repo).

This fixes #72

### DIFF
--- a/litex_boards/targets/orangecrab.py
+++ b/litex_boards/targets/orangecrab.py
@@ -85,10 +85,8 @@ class _CRG(Module):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCCore):
-    def __init__(self, sys_clk_freq=int(48e6), toolchain="trellis", **kwargs):
-        # Board Revision ---------------------------------------------------------------------------
-        revision = kwargs.get("revision", "0.2")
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @tommythorn! Merged.

PR #77, opened 2020-05-25
Take Ethernet PHY out of reset so default clock is 125 Mhz

skiphansen:
Without this change the terminal baud rate is 23,941 (measured by a scope) since the default clock is 25 Mhz instead of the expected 125 Mhz.  

The master clock on the Pano comes from the Ethernet PHY which is normally 125Mhz, but it is 25 Mhz while the PHY is in reset.  The active low reset pin is driven by FPGA pin R11. 

See https://github.com/tomverbeure/panologic-g2#fpga-external-clocking-architecture

### DIFF
--- a/litex_boards/platforms/pano_logic_g2.py
+++ b/litex_boards/platforms/pano_logic_g2.py
@@ -98,6 +98,7 @@ _io = [
         Subsignal("odt", Pins("J6"), IOStandard("SSTL18_II")),
     ),
     # Ethernet phy reset (clk125 is 25 Mhz instead of 125 Mhz if reset is active)
+    # See https://github.com/tomverbeure/panologic-g2#fpga-external-clocking-architecture
     ("gmii_rst_n", 0, Pins("R11"), IOStandard("LVCMOS33")),
 
 ]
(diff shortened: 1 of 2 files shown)


litex_boards/platforms/pano_logic_g2.py
```diff
@@ -97,6 +97,9 @@
         Subsignal("cke", Pins("D2"), IOStandard("SSTL18_II")),
         Subsignal("odt", Pins("J6"), IOStandard("SSTL18_II")),
     ),
+    # Ethernet phy reset (clk125 is 25 Mhz instead of 125 Mhz if reset is active)
```
mithro, line 100: Can you add a link to the documentation about this?

skiphansen: Done.

enjoy-digital:
Thanks! You were probably the first to test the target since i just prepared it and was planning to test it in the next days.

skiphansen:
Thanks Florent!

When you get around to the Pano perhaps you can add DDR support?  I'm struggling with https://github.com/timvideos/litex-buildenv/issues/444, I'm trying to merge DDR support from into litex-boards to see if the problem has been fixed recently, but I'm in way over my head with DDR.

enjoy-digital:
@skiphansen: the current target is minimal because i wanted to work on that and refactor/simplify the S6DDRPHY at the same time (https://github.com/enjoy-digital/litedram/issues/182). I should be able to work on it in the next days (but i'm not sure which revision of the board i have, i'll have to check).

skiphansen:
@enjoy-digital: Great, if you don't have a Rev C (or Rev B) I'd be more than happy to help you test, just let me know.

enjoy-digital:
@skiphansen: i just did a test on hardware and i have a Rev C, so i should be able to help you getting the DDR working on Rev C.

PR #78, opened 2020-05-27
ZCU104: add I2C

jedrzejboczar:
This adds I2C for the ZCU104 board, to be able to read SPD EEPROM from SO-DIMM modules.

### DIFF
--- a/litex_boards/platforms/zcu104.py
+++ b/litex_boards/platforms/zcu104.py
@@ -43,6 +43,12 @@ _io = [
         IOStandard("LVCMOS18")
     ),
 
+    ("i2c", 1,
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks, if this is the main (and only?) I2C bus of the board, we can proably just use 0 for the numbering: `("i2c", 0,...` and just do `platform.request("i2c")`. Can you change this if this is the case?

jedrzejboczar:
I did it because this I2C is referred to as I2C1 in the datasheet, but when looking more it seems that it is the only I2C available on this board, so I can change it.

enjoy-digital:
Thanks, merged.

PR #80, opened 2020-06-08
Use 128mb sdram, uart via i/o port on i/o board and vga terminal via i/o board

rob-ng15:
Change from 32mb to 128mb mister sdram, uart via usb-lookalike port on mister i/o board, vga via i/o board, 8 user leds, and light up the mister case leds

IMG_20200608_104115

IMG_20200607_142946

### DIFF
--- a/litex_boards/platforms/de10nano.py
+++ b/litex_boards/platforms/de10nano.py
@@ -18,6 +18,8 @@ _io = [
     ("user_led", 3, Pins("V15"),  IOStandard("3.3-V LVTTL")),
     ("user_led", 4, Pins("AF26"), IOStandard("3.3-V LVTTL")),
     ("user_led", 5, Pins("AE26"), IOStandard("3.3-V LVTTL")),
+    ("user_led", 6, Pins("Y16"), IOStandard("3.3-V LVTTL")),
+    ("user_led", 7, Pins("AA23"), IOStandard("3.3-V LVTTL")),
 
     ("key", 0, Pins("AH17"), IOStandard("3.3-V LVTTL")),
     ("key", 1, Pins("AH16"), IOStandard("3.3-V LVTTL")),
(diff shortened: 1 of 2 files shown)


rob-ng15:
For reference, the user i/o port on the mister i/o board pins AH9 and AG11 are Arduino i/o IO14 and IO15 respectedly.

rob-ng15:
https://github.com/rob-ng15/de10nano

have uploaded the bitstream to here

NOTE: SERIAL UART LOCATIONS

tx = AH9 Arduino i/o IO14
rx = AG11 Arduino i/o IO15

enjoy-digital:
Thanks, merged.

enjoy-digital:
I did some simplification with https://github.com/litex-hub/litex-boards/commit/85cac7abc0e1aaf4e5a92e3e93475e20958ce2d2 to make it more similar to others boards and ease maintenance. BTW the SDCard boot should now be working, could you do a test?

rob-ng15:
Will do.

rob-ng15:
Hi there,

Build fails as make.py has not been updated, should now reference BaseSoC.

Still not booting from sdcard in spi-mode, get boot.json file not found.
Tried with fat16, fat32 and exfat first partition. Any advice on what to
try next?

Rob.

enjoy-digital:
Hi @rob-ng15,

thanks for the feedback, i just updated make.py: https://github.com/litex-hub/linux-on-litex-vexriscv/commit/7d2bea65ff1af60f46be9fd207744fccd0bd4dbe.

For the SDCard, can you have a look at: https://github.com/enjoy-digital/litex/wiki/Load-Application-Code-To-CPU#sdcard-boot, you can use a FAT32 partition. It's everything is fine, can you try lowering the operational SPI clock: https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/liblitesdcard/spisdcard.c#L27 (IIRC you were using 12MHz on your setup) and eventually do a test with another SDCard (i know this is not the solution, but that's just to try to understand).

Sorry for the troubles, your solution was working correctly but the updated one will be more flexible and cover more use-cases, we just need to make sure it's working on the different boards. I already validated it on several boards but it's possible there are still a few things to fix.

Thanks,
 
Florent

rob-ng15:
Hi Florent,

I'll certainly try the reclocking and a different sdcard. I agree that this
fatfs solution is far more flexible.

Are there plans for a Linux driver for the terminal? When booted, it would
also be good for Linux to have access to the sdcard via spi, we could store
files on it then to make the system more fun!

Rob.

enjoy-digital:
That's indeed the next step for the SDCard, others have already had it working, but i still haven't tested it yet. It should only be a matter or enabling it in the .dts since we already have the driver. For the terminal, it would be able to use it for the LiteX console or for the Linux. This has already been done if you want to play with this: https://twitter.com/GregDavill/status/1231082623633543168

rob-ng15:
Hello Florent,

I'm up and running at 12500000 with a different sd card, not the one my
original tests were done on. So looking good, but not sure why no success
with my original card.

Rob.

PR #84, opened 2020-06-15
Change sdcard Pmod from JB to JD

mateusz-holenko:
This moves mappings of external SDCard Pmod to port JD, making it compatible with the the [Arty expansion board by Antmicro](https://github.com/antmicro/arty-expansion-board).


enjoy-digital:
Thanks, i also moved `spisdcard` to JD.

PR #87, opened 2020-07-07
arty: Add configuration of I2S pins

mateusz-holenko:


enjoy-digital:
Thanks, could you provide mode information about the extension that is used? Could it be added as a `_i2s_pmod_io` extension? (similar to what we did for the SD Card PMOD)

mateusz-holenko:
@enjoy-digital You are right - we should be more specific. 

The commit contains now the information about the PMOD model + pins configuration is moved to a separate `_i2s_pmod_io` array so it can be enabled using `self.platform.add_extension(arty_platform._i2s_pmod_io)` when needed.

enjoy-digital:
Thanks for the update, merged.

PR #88, opened 2020-07-09
OrangeCrab: Update platform

gregdavill:
I was building the foboot bootloader for both r0.2 and r0.1 boards, and realised that I was missing some io definitions on the r0.1 variants.

They're now updated. 
Along with fixing the standard extensions.
Changing the VID/PID pair from the fomu to our own: http://pid.codes/1209/5AF0/


enjoy-digital:
Thanks!

PR #94, opened 2020-08-01
switch ULX3S to OpenFPGALoader

pepijndevos:
The ULX3S target currently uses the board-specific ujprog, which does not compile out of the box for me.
This PR changes it to use openFPGALoader, which is a more generic and broadly supported option.
This PR depends on https://github.com/enjoy-digital/litex/pull/615

### DIFF
--- a/litex_boards/platforms/ulx3s.py
+++ b/litex_boards/platforms/ulx3s.py
@@ -3,7 +3,7 @@
 
 from litex.build.generic_platform import *
 from litex.build.lattice import LatticePlatform
-from litex.build.lattice.programmer import UJProg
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks for the PR, unfortunately i have trouble using openFPGALoader on my config (does not work and not sure i have time understanding why) and since the ULX3S team provide a minimal programmer for their board, that is small and has always been reliable, i think it makes sense to keep if the default programmer. Since https://github.com/enjoy-digital/litex/pull/615 is merged, it will be easier to use openFPGALoader in your specialized target file.

PR #103, opened 2020-08-24
Adding initial support for the Crosslink-NX

davidcorrigan714:
Testing for all of this has been with the Crosslink-NX evaluation board, though it should extend to all of Lattice's Nexus family chips. There's one or two TODOs in here to account for variations in device features. There is also some discrepancy on which chip is on the evaluation board, some seem to have -8 speed grade chips.


enjoy-digital:
Thanks, this has been rebased/merged manually with https://github.com/litex-hub/litex-boards/commit/795e34aafd15868832df1611d82529ac70f3c089 (and also adapted to the renaming done in LiteX).

PR #104, opened 2020-08-27
Add support for 5A-75E V6.0 board

DerFetzer:
...though it is not working with Ethernet for me, yet.

When I try current consumption goes way down and nothing happens. Except for I can probe a weird 2.5MHz clock at the SDRAM.
Does anybody have an idea?

### DIFF
--- a/litex_boards/platforms/colorlight_5a_75e.py
+++ b/litex_boards/platforms/colorlight_5a_75e.py
@@ -4,8 +4,9 @@
 # Copyright (c) 2020 Vadim Kaushan <admin@disasm.info>
 # SPDX-License-Identifier: BSD-2-Clause
 
-# The Colorlight 5A-75E PCB and IOs have been documented by @derekmulcahy:
+# The Colorlight 5A-75E PCB and IOs have been documented by @derekmulcahy and @adamgreig:
 # https://github.com/q3k/chubby75/issues/59
+# https://github.com/q3k/chubby75/pull/67
 
 from litex.build.generic_platform import *
 from litex.build.lattice import LatticePlatform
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks, for Ethernet, you could maybe do a test by commenting out the` rst_n` definition in the Platform file and use the 25MHz from Ethernet for the SoC.

DerFetzer:
Yes, you are right. Problem is that the 25MHz clock is generated by one of the Ethernet PHYs so it stops on PHY reset.
I added the possibility to use the ECP5 OSCG primitive for system clock.

DerFetzer:
This is working as well. I tried that yesterday or so.
Question is what would be the preferred thing to do?
I tried removing the `rst_n` pin programmatically so you could provide a command line switch. But it did not work due to some python magic in the framework.

enjoy-digital:
@DerFetzer ah ok, so Ethernet is working when the internal oscillator is used? Sorry i thought it was not the case. If so, we could merge and i will try to find a better solution in the next days.

DerFetzer:
Yes, Ethernet is working. 
I did not do a full test, yet. But when I use `netboot` command from `LiteX BIOS` I can see `ARP` packets coming from the board.

DerFetzer:
I seems like only transmitting over Ethernet is possible. Though my PC answers to the `ARP` requests there are no other packets coming from the board. This is the same with the `rst_n` pin commented out and with external clock.

The only obvious problem I see might be the same as in https://github.com/enjoy-digital/liteeth/issues/47. 
But the timing for my board is even worse:
```
Warning: Max frequency for clock '$glbnet$eth_clocks_rx$TRELLIS_IO_IN': 87.94 MHz (FAIL at 125.00 MHz)
```
In the issue you wrote that this might not be a problem but I don't know if this statement still stands with such a big timing violation.

DerFetzer:
When I limit the Ethernet link to 100MBit/s or use a 100MBit/s switch there are not even ARP frames coming through. This is strange.

DerFetzer:
GBit-Ethernet is working now! There is a pin swap at the RXD pins for PHY0 in the [documenation](https://github.com/q3k/chubby75/blob/master/5a-75e/hardware_V6.0.md). I will make a pull request there.

The problem with 100MBit/s persists though.

enjoy-digital:
Thanks for the update, i think we can merge now. The ECP5RGMII Ethernet PHY hasn't been test/validated at 100Mbps so that's probably not related to the board.

PR #106, opened 2020-09-04
alveo_u250: Add PCIe x4 support

daveshah1:
Based on the implementation in xcu1525.

Note that enjoy-digital/litepcie#39 needs to be fixed before this builds out-of-the-box.

scratch_test seems to work correctly:
```
$ ./litepcie_util scratch_test
Write 0x12345678 to scratch register:
Read: 0x12345678
Write 0xdeadbeef to scratch register:
Read: 0xdeadbeef
```
I haven't tested any further than that. I also found I needed to reboot for the device to appear correctly, but my setup is somewhat unusual as it involves Thunderbolt on a X399 motherboard without the proper BIOS support.

### DIFF
--- a/litex_boards/platforms/alveo_u250.py
+++ b/litex_boards/platforms/alveo_u250.py
@@ -275,6 +275,17 @@ _io = [
             "AP7 AR9 AT7 AU9 AV7 BB5 BD5 BF5")),
     ),
 
+    # pcie
+    ("pcie_x4", 0,
+        Subsignal("rst_n", Pins("BD21"), IOStandard("LVCMOS12")),
+        Subsignal("clk_n", Pins("AM10")),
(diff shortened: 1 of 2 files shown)


daveshah1:
> I also found I needed to reboot for the device to appear correctly, but my setup is somewhat unusual as it involves Thunderbolt on a X399 motherboard without the proper BIOS support.

The solution to this is almost https://github.com/enjoy-digital/litepcie/issues/36#issuecomment-652849400, but in my case I have to direct the rescan just to the Alpine Ridge device and not the whole PCI subsystem; or my machine totally locks up.

enjoy-digital:
Thanks, merged. The naming issue with be fixed soon and more testing will be done in the next days with LitePCIe/Ultrascale+ devices.

PR #109, opened 2020-09-25
orangecrab: Fix r0.1 user_led mapping

geertu:
On r0.1, all three user_leds are mapped to the same pin.
Fix this by mapping them to the pins connected to the individual
channels of the RGB LED, to match the comments, the schematics, and the
spirit of r0.2.

Untested on real hardware (I have r0.2 only).

Fixes: c94cbae0c04bbda3 ("orangecrab: add user_led (RGB leds), DFUProg and --load support.")
Signed-off-by: Geert Uytterhoeven <geert@linux-m68k.org>


enjoy-digital:
Thanks.

PR #110, opened 2020-09-30
Add initial support for Trenz TEC0117 board

pepijndevos:
This PR complements https://github.com/enjoy-digital/litex/pull/654 and adds support for the TEC0117 board.

Like the Icebreaker target, it puts the BIOS in flash, but since this board has internal bitstream flash and external user flash that is not accessible through JTAG, some hacky stuff is required to program the flash.

It currently uses 2kB sram, the minimal the bios was happy with. That's 2 bram primitives out of 26.

### DIFF
--- a/litex_boards/platforms/tec0117.py
+++ b/litex_boards/platforms/tec0117.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+from migen import *
+
+from litex.build.generic_platform import *
+from litex.build.gowin.platform import GowinPlatform
+from litex.build.openfpgaloader import OpenFPGALoader
+
+# IOs ----------------------------------------------------------------------------------------------
+
+_io = [
+    ("user_led",  0, Pins("86"), IOStandard("LVCMOS33")),
+    ("clk12", 0, Pins("35"), IOStandard("LVCMOS33")),
+    ("rst",   0, Pins("77"), IOStandard("LVCMOS33")),
+
+    ("user_led", 0, Pins("86"), IOStandard("LVCMOS33")),
+    ("user_led", 1, Pins("85"), IOStandard("LVCMOS33")),
+    ("user_led", 2, Pins("84"), IOStandard("LVCMOS33")),
+    ("user_led", 3, Pins("83"), IOStandard("LVCMOS33")),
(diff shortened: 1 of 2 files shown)


pepijndevos:
So the current state of things is that I can run the bios with serv and picorv32, but with vexrisc it hangs on the ascii art and copyright and does never get to the CRC check or the prompt. I'm not sure if this is a synthesis bug, it's running out of memory, UART is broken, or something else.

As far as running my own (Rust) code goes, I have not yet had success. This is probably my own fault.
By default lxterm tries to write to main_ram, which this target has none of. What's odd is that for `simple.py` it just writes into the void and hangs, while on this target it just hangs at 0%.
I did try writing to sram, which for `simple.py` hangs at 12% (presumably because it overwrites the BIOS), and for this target is similarly stuck at 0%.
I have not "bisected" the differences between `simple.py` and this target, but I assume for correct operation I have to compile and load my code in some RAM that actually exists and is not used by the BIOS.

Note that there seems to be a bug with OpenFPGALoader when it comes to writing to flash. And there is another bug in libftdi that does not reattach the UART after upload. My current hacky upload process is to run
```
python litex_boards/targets/tec0117.py --cpu-type picorv32 --cpu-variant minimal --csr-svd trenz.svd --csr-data-width 32 --build --flash // no --load
sudo ~/bin/gowin1.9.6/Programmer/bin/programmer_cli --device GW1NR-9 --run 6 --fsFile /home/pepijn/code/litex_workspace/litex-boards/build/tec0117/gateware/impl/pnr/project.fs
```
This uploads the BIOS to the user flash, and then uses the vendor programmer to write the bitstream to the internal flash. This leaves the UART inoperable, so you have to unplug and replug the board for `/dev/ttyUSB1` to be usable with `lxterm`.

@trabucayre has been very responsive to fixing his awesome tool, which already includes a workaround that reattaches the UART. So once the flash bug is fixed, it will be possible to upload the BIOS and bitstream with one command, and use the UART right away. (I think he's also considering native SPI programming support, removing the need for the PySpiFlash workaround)

enjoy-digital:
Thanks, this is merged. If you were using the Minimal variant of VexRiscv, it's possible the issue was: https://github.com/enjoy-digital/litex/pull/663.

PR #112, opened 2020-10-06
Zedboard

yetifrisstlama:
Adds a platform file for the Zedboard. Including FMC connector definition.

### DIFF
--- a/litex_boards/platforms/zedboard.py
+++ b/litex_boards/platforms/zedboard.py
@@ -1,0 +1,256 @@
+# This file is Copyright (c) 2019 Michael Betz <michibetz@gmail.com>
+# License: BSD
+
+from litex.build.generic_platform import Pins, IOStandard, Subsignal
+from litex.build.xilinx import XilinxPlatform, XC3SProg, VivadoProgrammer
+from litex.build.openocd import OpenOCD
+from migen import *
+
+_io = [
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks, this is merged. I did some changes with https://github.com/litex-hub/litex-boards/commit/55da8b867a2f04e53df1f17b61d327354dffcd1e to uniformize it a little bit more with the other platforms. Could you also provide the `digilent_zedboard.cfg`  OpenOCD configuration file you are using? 

Also for the PS7 IOs, i'm not a heavy user of Zynqs, but i'm not even sure the PS7 locations need to be defined in the target and think it can be similar to https://github.com/litex-hub/litex-boards/blob/master/litex_boards/platforms/zybo_z7.py#L42-L67. Could you check this? If this is not required, we could just use the simplified definition.

yetifrisstlama:
Thanks for merging and making the syntax more uniform to the other files.

> Could you also provide the digilent_zedboard.cfg OpenOCD configuration file you are using
It's the one included in openocd since 2014:
https://repo.or.cz/openocd.git/blob_plain/HEAD:/tcl/board/digilent_zedboard.cfg

> If this is not required, we could just use the simplified definition.
Technically it works without the explicit definition. I've added them though, to get rid of a bunch of critical warnings in vivado.

PR #113, opened 2020-10-08
versa_ecp5: Add etherbone support

kbeckmann:
Etherbone can be enabled with --with-etherbone


enjoy-digital:
Thanks @kbeckmann! Merged.

PR #114, opened 2020-10-09
versa_ecp5: Add --eth-phy to select ethernet phy

kbeckmann:
This also simplifies the logic a bit.


kbeckmann:
Should have included this in the previous patch yesterday, but at least now it's the same as colorlight_5a_75x.

enjoy-digital:
Thanks! Merged.

PR #115, opened 2020-10-11
platforms/ulx3s: Update ULX3S SD pins for revision 2.0

BryanJacobs:
The ULX3S has three revisions, with three different sets of FPGA pins used for the SD card.

The 2.0 revision has just shipped in relatively large quantities from a successful CrowdSupply campaign, and is likely to be the most popular version going forward. This change updates the litex platform file to correctly wire the SD card, allowing boot-from-SD in the litex BIOS on a v2.0 ULX3S board.

See https://raw.githubusercontent.com/emard/ulx3s/fe78a65e7cf8c5eba9c175209471edcc5004f67c/doc/schematics_v312.pdf , page 6, to confirm the pinout and see the changes between board revisions.

I am uncertain how the litex project wishes to handle different board hardware variations (perhaps adding a "version" kwarg?) but at least the default should probably match the 2.0 board...


enjoy-digital:
Thanks for the PR and feedback on boot-from-SD card with the 2.0 revision. The initial LiteX target has been developed with the 1.7 revision and i agree with you that it should default to 2.0. I just merged you PR and will add back 1.7 revision support with an additional argument.

enjoy-digital:
I just added support for both revisions with https://github.com/litex-hub/litex-boards/commit/ddf7038c78f5f7eff5d2103c2e39821459504734.

PR #117, opened 2020-10-17
Initial support for the MIST board (https://github.com/mist-devel/mis…

YanekJ:
…t-board/wiki)


enjoy-digital:
Thanks at lot for your feedback (by mail) and contribution! 

Here is the boot screenshot sent by @YanekJ: 
image

PR #120, opened 2020-11-02
Add LFE5U-12F device for ULX3S

kevinmehall:
The ULX3S board is also available with the 12F size of the ECP5.


enjoy-digital:
Thanks, merged.

Issue #42, opened 2020-02-16
boot hadbadge from flash


pdp7:
While the kernel boots in under 30 seconds, it [currently it takes about 2 minutes](https://asciinema.org/a/293450) to load the kernel Image and rootfs.cpio:
```
lxterm --images=images.json /dev/ttyUSB0 --speed=1e6 --no-crc
```

![Screenshot from 2020-01-30 19-44-10](https://user-images.githubusercontent.com/313864/73479783-e8520200-4398-11ea-8192-abe5f5bdfad5.png)

The badge and cartridge both have:
- [Winbond W25Q128JVSIQ 128 Mbit NOR flash chip](https://www.digikey.com/product-detail/en/W25Q128JVSIQ/W25Q128JVSIQ-ND/5803943/?itemSeq=316234062)

I want to use the built-in Flash to store the kernel and rootfs rather than load over serial.

I initially was discussing this in #35 but I wanted to start a new issue to decouple this effort from the memory access optimization that @enjoy-digital already completed in enjoy-digital/litedram@721c84b and enjoy-digital/litex@39ce39a

pdp7:
I'm trying to get the Hackaday Badge to boot from the SPI flash as it takes 2 mintues to load the kernel and rootfs over serial.

Repo version:
```
pdp7@x1:~/dev/enjoy/linux-on-litex-vexriscv$ git log --oneline -1
cab81d5 (HEAD -> master, origin/master, origin/HEAD) sim: use new sdram verbosity parameter
```
Repo diff:
```diff
diff --git a/soc_linux.py b/soc_linux.py
index ecf875b..34105c0 100644
--- a/soc_linux.py
+++ b/soc_linux.py
@@ -104,7 +104,7 @@ def SoCLinux(soc_cls, **kwargs):
 
         def add_spi_flash(self, dummy_cycles):
             # TODO: add spiflash1x support
-            spiflash_pads = self.platform.request("spiflash4x")
+            spiflash_pads = self.platform.request("spiflash")
             self.submodules.spiflash = SpiFlash(
                 spiflash_pads,
                 dummy        = dummy_cycles,

```


From `linux-on-litex-vexriscv/soc_linux.py`:
```        mem_map = {**soc_cls.mem_map, **{
            "emulator_ram": 0x20000000,
            "ethmac":       0xb0000000,
            "spiflash":     0xd0000000,
            "csr":          0xf0000000,
        }}
```
[...]

PR #125, opened 2020-11-23
Add support for proFPGA XCVU19P board

piotr-binkowski:
This PR aims to add support for proFPGA XCVU19P board (https://www.profpga.com/profpga-xcvu19p).
Before merging serial port configuration needs to be adjusted since current pins are connected to onboard LEDs (and it looks like this board has no onboard USB-UART converter).
@enjoy-digital what would be the best way to handle UART in this case?
Is is possible to disable it entirely or maybe configure it to be accessible via PCIe-Wishbone bridge?

Also this platform represents only one configuration of this board (FPGA board with PCIe adapter attached to port TA1) but in theory many more modules can be attached to different expansion ports so I'm wondering if there is a better way to represent such configurations in Litex?

### DIFF
--- a/litex_boards/platforms/profpga_xcvu19p.py
+++ b/litex_boards/platforms/profpga_xcvu19p.py
@@ -1,0 +1,74 @@
+#
+# This file is part of LiteX-Boards.
+#
+# Copyright (c) 2020 David Shah <dave@ds0.me>
+# Copyright (c) 2020 Florent Kermarrec <florent@enjoy-digital.fr>
+# Copyright (c) 2020 Antmicro <www.antmicro.com>
+# SPDX-License-Identifier: BSD-2-Clause
+
+from litex.build.generic_platform import Pins, Subsignal, IOStandard, Misc
+from litex.build.xilinx import XilinxPlatform, VivadoProgrammer
+
+# IOs ----------------------------------------------------------------------------------------------
+
+_io = [
+    # CLK1
+    ("clk300", 0,
+        Subsignal("n", Pins("CA40"), IOStandard("LVDS")),
+        Subsignal("p", Pins("CA39"), IOStandard("LVDS")),
+    ),
+    # Serial
+    ("serial", 0,
+        Subsignal("rx", Pins("BE29"), IOStandard("LVCMOS18")),
+        Subsignal("tx", Pins("BE30"), IOStandard("LVCMOS18")),
+    ),
+    # EB-PDS-PCIe-Cable-R3 on TA1
+    ("pcie_x4", 0,
+        Subsignal("rst_n", Pins("E40"), IOStandard("LVCMOS12")),
+        Subsignal("clk_n", Pins("AU10")),
+        Subsignal("clk_p", Pins("AU11")),
+        Subsignal("rx_n",  Pins("AJ1 AK3 AL1 AM3")),
+        Subsignal("rx_p",  Pins("AJ2 AK4 AL2 AM4")),
+        Subsignal("tx_n",  Pins("AL6 AM8 AN6 AP8")),
+        Subsignal("tx_p",  Pins("AL7 AM9 AN7 AP9")),
+    ),
+    ("pcie_x8", 0,
+        Subsignal("rst_n", Pins("E40"), IOStandard("LVCMOS12")),
+        Subsignal("clk_n", Pins("AU10")),
+        Subsignal("clk_p", Pins("AU11")),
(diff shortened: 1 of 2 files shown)


enjoy-digital (changes requested):
Thanks @piotr-binkowski,

- for the serial, you can remove it and just force the uart to `crossover` (kwargs["uart_name"] = crossover).
- for the PCIe/Connector, you should probably create connectors for the different connectors of the boards (ex: https://github.com/litex-hub/litex-boards/blob/master/litex_boards/platforms/arty.py#L158) and then create functions for the adapter you can plug on a connector, ass done for example here: https://github.com/litex-hub/litex-boards/blob/master/litex_boards/platforms/arty.py#L285-L307.
- for LitePCIe, can you use the new LiteX's add_pcie method to simplify the integration: https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/xcu1525.py#L93-L99

Thanks

mithro:
FYI - The person writing this is @piotr-binkowski not @esden :-P -- Too many Piotr's I think :-)

enjoy-digital:
Ah sorry, i was aware i was answering to @piotr-binkowski but auto-completion probably suggested the wrong Piotr...

enjoy-digital:
@piotr-binkowski, @mithro: after thinking a bit about it, I don't think it's a good idea to integrate this proFPGA VU19P: most of the boards used in LiteX and the ecosystem have been selected to be useful and affordable to allow others to reproduce results. The Ultrascale boards are still expensive but I tried to select affordable ones (BCU1525 and FK33 that can be found around 1k$), which still allow some users of the community to buy them and play with them. Here with the proFPGA VU19P we are probably around 70k$ the board (and probably more), so this is only accessible to very large companies.

I'm already using proFPGA boards for some clients and testing the cores on them, but maintain them in specific repositories. So this configuration should probably be supported and maintained in a specific project/repository. The current Ultrascale(+) code is also working, but some refactoring is needed to ease supporting more boards, for example: #42. So let's first finish things that have been started before going further.

mithro:
@enjoy-digital - I don't see any reason to not have all boards which people are using in this repo? Otherwise multiple people will be writing these and getting them wrong.

mithro:
Just like Linux supports extremely large 100 CPUs IBM, $100k USD, systems and tiny little 1 core, $10 USD RPi Zero systems, everyone benefits by working together and sharing their work in common place. 

Also, yesterday's $10k USD system is tomorrow's $1k USD system. Today, the $35 USD RPi is a Quad core Cortex-A72 (ARM v8) 64-bit SoC @ 1.5GHz beast, while the first RPi had a single-core 700MHz CPU :-).

enjoy-digital:
The proFPGA boards are not classical FPGA board similar to ones supported in LiteX-boards but prototyping systems. They are basically just FPGAs exposing IOs on connectors and that you can customize by adding mezzanine adapter boards. So to support them correctly, for each prototyping system, you need to create a platform file, then create the adapter files (for the different mezzanines that you can plug on it) that you can reuse between platforms and then example targets for the different possible configurations (with PCIe mezzanine, DDR4 mezzanine, etc....) So in addition to being very expensive, correctly supporting these boards is a whole project by itself.

So this is not the aim of this repository. Creating an external project for proFPGA boards with plarforms/adapters/examples could be interesting, but I don't have the resources/time to maintain it.

enjoy-digital:
@piotr-binkowski, @mithro: the work from this PR and from https://github.com/enjoy-digital/litepcie/pull/44 has been integrated in https://github.com/litex-hub/litex-on-profpga-systems with additional simplifications and features. I'll go a bit further after lunch but we can continue the discussion here if you want. I'm just trying to create a skeleton to ease supporting proFPGA boards, this could be used for this PoC, and could be extended later if more proFPGA boards are supported but I don't think I'll have the time/ressources to maintain it.

enjoy-digital:
https://github.com/litex-hub/litex-on-profpga-systems has been updated with an initial attempt to implement the ideas described in the previous comments. I also listed ideas in issues that I think could be interesting to implement if we want to explore more with proFPGA boards and LiteX. 

Sorry if closing the issue has been miss-interpreted, I was doing a bunch of update on issues/PR of the project and maybe closed this one to fast. I'll be more careful in the future but would also like similar PRs to this be discussed in issues first, from a maintainer point of view I find easier to say say no to an idea in an issue than in a PR since I'm well aware that some work has already been done in the PR and closing it could generate frustration.

Anyway, leaving aside these integration aspects, I'm very happy to see LitePCIe validated and used on such hardware.

PR #127, opened 2020-11-25
nexus: Allow selection of toolchain

daveshah1:
Please see the caveats in enjoy-digital/litex#709 - for the reasons mentioned there the only working combo ATM is VIP+hyperram


enjoy-digital:
Thanks, that's merged.

PR #128, opened 2020-11-26
add support for redpitaya14/16

trabucayre:
[redpitaya](https://www.redpitaya.com/) are zynq based boards with ADC and DAC.
This PR provides support for two variant:
- redpitaya/STEMlab 125-14: zynq xc7z010clg400 with 14 bits ADC & DAC @ 125MS/s
- redpitaya/SDRlab 122-16: zynq xc7z020clg400 with 16 bits ADC & 14 bits DAC @ 122MS/s

This PR allows to use the ARM cortex A9.

### DIFF
--- a/litex_boards/platforms/redpitaya.py
+++ b/litex_boards/platforms/redpitaya.py
@@ -1,0 +1,167 @@
+#
+# This file is part of LiteX-Boards.
+#
+# Copyright (c) 2020 Gwenhael Goavec-Merou <gwenhael.goavec-merou@trabucayre.com>
+# SPDX-License-Identifier: BSD-2-Clause
+
+from litex.build.generic_platform import *
+from litex.build.xilinx import XilinxPlatform, VivadoProgrammer
+
+# IOs ----------------------------------------------------------------------------------------------
+
+_io = [
+    # Leds
+    ("user_led", 0, Pins("F16"), IOStandard("LVCMOS33")),
+    ("user_led", 1, Pins("F17"), IOStandard("LVCMOS33")),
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @trabucayre! This looks good. As you saw the Zynq7000 .xci integration is still a WIP, thanks for doing the same than on others Zynq7000 boards, we'll improve it in the future.

trabucayre:
Thanks!
For xci I search a clean way too.
One possible solution is to use block design and preset to generate this file (for some boards, supported by vivado, it's quite easy, in other case a dict with around 50lines must be added somewhere (maybe in platform file)).
Anyway using an xci imply a imutable configuration without a way to update this according to the targeted projet (Axi HP, interrupts, ...), with a dict is always possible to append default parameters at the target level (but It's add noise in  platform or target .py)

enjoy-digital:
Indeed I'd also like to find a convenient/flexible solution, that's why i was just reusing an external .xci for now as an intermediate step. One solution could be to directly instantiate the PS7 primitive in the design and do the equivalent of the .xci around it (as done in the Axiom camera for example: https://github.com/apertus-open-source-cinema/axiom-firmware/blob/master/peripherals/soc_main/ps7_stub.vhd#L373)

trabucayre:
Yes. I have already seen this approach. But with this, we are losing ability to have `ps7_gpl_init` file (used by `u-boot` to produces the `fsbl`) with the corresponding configuration .
Okay, In most case, default `u-boot`'s files are enough and some configuration may be done at `u-boot` and/or `linux` level, but it's sometime useful to have access to this file.

PR #130, opened 2020-12-07
zybo_z7: fix clock pin constraint

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

This PR fixes an issue with the pin mapping of the Zybo Z7 board.

In fact, the external clock pin is tied to an incorrect pin, which was valid fro the Zybo board, but not for the Z7 version.


enjoy-digital (approved):
Thanks @acomodi.

PR #134, opened 2020-12-19
Fix FPGA reset logic for orangecrab target

Disasm:
This PR fixes FPGA reset logic in `_CRG` to match one in `_CRGSDRAM`. Without this fix FPGA SRAM programming fails when SDRAM is not used.

The bug was introduced in https://github.com/litex-hub/litex-boards/commit/9b6ed6bdf1e2b16400db56900def800d8ef3ee78.


enjoy-digital:
Thanks @Disasm.

PR #136, opened 2020-12-22
nexys4ddr: add support for litexvideo VGA Terminal

devobliquezer0:
This commit adds VGA support for the Nexys A7/ Nexys 4 DDR.

The VGA is however limited to RGB443 instead of the full 12bit RGB444.
This is because IO D8 which is MSB for Blue, is also used for ETH int_n.
This makes the final output have a yellow tint.

### DIFF
--- a/litex_boards/platforms/nexys4ddr.py
+++ b/litex_boards/platforms/nexys4ddr.py
@@ -129,6 +129,16 @@ _io = [
         Subsignal("int_n",   Pins("D8")),
         IOStandard("LVCMOS33")
      ),
+
+    # VGA
+     ("vga", 0,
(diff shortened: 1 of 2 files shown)


devobliquezer0:
I'm curious if `region=SoCRegion(origin=0x30000000, size=0x10000)` is required for Xilinx platforms, I just took that block from de10nano.

enjoy-digital:
Thanks @devobliquezer0, it's merged and with https://github.com/litex-hub/litex-boards/commit/9beaf25822c3794e14e8773778e014b3d7d9c80b I fixed the `eth.int_n ` pin and enabled the 4-bit on `vga.blue`.

PR #138, opened 2020-12-26
Readme Fixed ECPIX-5 Sys-Clk frequency to 100MHz

bvernoux:
Fixed ECPIX-5 Sys-Clk frequency to 100MHz (instead of 75 MHz) see http://docs.lambdaconcept.com/ecpix-5/features/clock-reset.html


bvernoux:
It will be great to clarify that in factor native clock input is 100MHz but the SOC Clock (to be clarified) is set to 75MHz ?

enjoy-digital:
This has been clarified in https://github.com/litex-hub/litex-boards/issues/139.

PR #140, opened 2020-12-27
acorn_cle_215: Fix hard-lockup on PCIe bringup.

tomparks:
When building the SQRL CLE 215/nitefury target, the host crashes after programming.

Repro steps: 
1) clone current litex
2) python upstream/litex-boards/litex_boards/targets/acorn_cle_215.py --uart-name=crossover --with-pcie --build --driver
3) flash with vivado hw manager (I'm using the platform cable, so I flash with the vendor tools)
4) execute `lspci`, or otherwise rescan the bus
5) system locks up: display freeze, ssh sessions die, does not respond to ping or power button, etc. 

No logs in dmesg make it to disk, and I havn't got a UART cable to see if there's any kernel logging that gets displayed. I've reproduced on a Intel (Intel Xeon CPU E3-1240) and AMD (AM1) platform, both Ubuntu 20.04.

I believe this is due to the PCIe bus width being set to 128b. Running git bisect narrows down the issue to a commit in litepcie - with eveything else up to date with HEAD:

```
# first bad commit: [899a0e9d68ddb58d7b231f25332216692c8dd601] phy/s7pciephy/uspciephy: set default pcie_data_width to None and set it to data_width if None.
```

manually specifing the old default bus width (64b) allows the device to be enumerated correctly and the driver info & DMA tests work fine. Is this a reasonable fix & if so, 1) why does 128b break and should this be replicated for other PCIe targets?

### DIFF
--- a/litex_boards/targets/acorn_cle_215.py
+++ b/litex_boards/targets/acorn_cle_215.py
@@ -105,7 +105,7 @@ class BaseSoC(SoCCore):
         # PCIe -------------------------------------------------------------------------------------
         if with_pcie:
             self.submodules.pcie_phy = S7PCIEPHY(platform, platform.request("pcie_x4"),
-                data_width = 128,
+                data_width = 128, pcie_data_width=64,
                 bar0_size  = 0x20000)
             self.add_csr("pcie_phy")
             self.add_pcie(phy=self.pcie_phy, ndmas=1)


shawnl:
Is the platform cable the only way to program the acorn 215+? I have the board, but no platform cable.

enjoy-digital:
Thanks @tomparks for the investigation and sorry for this stupid issue. In fact I think we should fix things the other way around and use a 128-bit Datapath on the PHY as I just did in https://github.com/enjoy-digital/litepcie/commit/2f8197170aa122f3fe40e97c948133901902cd42. 

In some cases, it's useful to use different data-width between the PHY and LitePCIe core, but for the LiteX-Boards examples this is not useful and it's probably better to use similar data-width. This was not working since the PHY was expecting a 64-bit data-width while the core was configured to 128-bit. We could detect this issue earlier but this would requires reading the contents of the `.xci` file which we tried to avoid for now.

Can you do a test with the attached bitstream that I just re-generated with https://github.com/enjoy-digital/litepcie/commit/2f8197170aa122f3fe40e97c948133901902cd42 applied? (
` ./acorn_cle_215.py --uart-name=crossover --with-pcie --build`): [acorn_cle_215_with_pcie_uart_name_crossover_2020_12_28.zip](https://github.com/litex-hub/litex-boards/files/5748119/acorn_cle_215_with_pcie_uart_name_crossover_2020_12_28.zip)

enjoy-digital:
@shawnl: I'm sure it's possible to update the SQRL bitstream over PCIe but we would need to spend time understand the protocol which hasn't been done yet. In the LiteX bitstream you will also need to integrate the logic to access the Flash to be able update it later (which is possible and available in LiteX but hasn't been integrated in the Acorn CLE215+ example). So avoiding using a JTAG programmer should be possible but would requires some work.

tomparks:
> Can you do a test with the attached bitstream that I just re-generated with [enjoy-digital/litepcie@2f81971](https://github.com/enjoy-digital/litepcie/commit/2f8197170aa122f3fe40e97c948133901902cd42) applied? 

The attached bitstream works fine; ident is `FPGA identification: LiteX SoC on Acorn CLE 215+ 2020-12-28 17:32:57` and the utils tests work. Thank you for the rapid fix & feel free to close the PR :)

enjoy-digital:
Great, thanks @tomparks for the feedback. Sorry for this, I hope the time you spent on it was also a way to learn a bit about LiteX/LitePCIe internals.

PR #144, opened 2021-01-04
genesys2: LiteSDCard support

gsomlo:
Platform patch based on [this entry from the genesys2 board manual](https://reference.digilentinc.com/reference/programmable-logic/genesys-2/reference-manual#microsd_slot). Additionally, some minor cleanup modelled after the equivalent entries in the `nexys4ddr` platform file.

***EDIT***: added commit to support `--with-[spi-]sdcard` on the target build command line


enjoy-digital:
Thanks @gsomlo, merged.

PR #145, opened 2021-01-05
colorlight: Add option for etherbone ip address and LED chaser

hansfbaier:
Hi, I added an option to the colorlight board to customize the etherbone ip adress.
I also made the LED chaser optional on the colorlight board, because
the uart is enabled by default which uses the same pin as the LED.

### DIFF
--- a/litex_boards/targets/colorlight_5a_75x.py
+++ b/litex_boards/targets/colorlight_5a_75x.py
@@ -119,7 +119,7 @@ class _CRG(Module):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCCore):
-    def __init__(self, board, revision, sys_clk_freq=60e6, with_ethernet=False, with_etherbone=False, eth_phy=0, use_internal_osc=False, sdram_rate="1:1", **kwargs):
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @hansfbaier, I'll have a closer look at this in the next days: I'd like all targets to expose the same options.

hansfbaier:
@enjoy-digital : I understand the desire to be consistent, but since all boards have different hardware options, I would not see a problem here for the software to model those. OTOH, I already now am used to subclass boards to customize them, and  this is probably the intended design. But I really think the IP addresses should be customizable out of the box (in my example my home network was already 192.168.1.x so I needed to reconfigure)

enjoy-digital:
@hansfbaier: sure I understand, I just wanted to also apply this changes to the others targets with Etherbone. I'm going to merge it now and will do the other changes later.

hansfbaier:
@enjoy-digital I can apply it to the other targets also, so you won't have to (can do that tomorrow).

enjoy-digital:
This is merged, I just added https://github.com/litex-hub/litex-boards/commit/c829a47c31777dda79a787af7277133cdf7bae96 to automatically disable the LedChaser when the serial is used (to avoid the `with_leds` argument since this behaviour is a bit specific to hardware limitations with this board).

I also renamed `etherbone-ip` to `eth-ip` to also use it in the future to configure the IP Address when using `--with-ethernet`.

Applying the `eth-ip` parameter to the other targets would be welcome, but don't feel obligated.

hansfbaier:
@enjoy-digital OK I will do that tomorrow too. Thanks!

PR #146, opened 2021-01-07
add etherbone ip address option for all boards with etherbone

hansfbaier:
Hi,
I added the eth-ip option to all boards with etherbone.
I also tested with a debugger that the option value arrives in LiteXSoC.
During those tests I noticed that on the sds1104xe board
I get an exception, because with_uart is active while the platform defines
no uart pins, so I disabled it in the commit below.
I am not perfectly happy with this solution, because when I tried to
add `with_uart` options to the constuctor calls of the SoC, I
got double keyword exceptions. So something I could not
find out in reasonable time automatically set this option.

### DIFF
--- a/litex_boards/targets/arty.py
+++ b/litex_boards/targets/arty.py
@@ -56,7 +56,7 @@ class _CRG(Module):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCCore):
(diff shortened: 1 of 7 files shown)


enjoy-digital:
Thanks, I pushed your changed manually. For the `sdr1104xe` it's because the crossover UART has to be used (https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/sds1104xe.py#L11). I'll update the target to make it the default.

PR #151, opened 2021-01-14
Arty: add dbg periph for dbg CPU variant

kgugala:
This PR adds debug logic to the design if a debug CPU variant is requested. The code will automatically map the debug peripheral on the bus.

### DIFF
--- a/litex_boards/targets/arty.py
+++ b/litex_boards/targets/arty.py
@@ -102,6 +102,11 @@ class BaseSoC(SoCCore):
             sys_clk_freq = sys_clk_freq)
         self.add_csr("leds")
 
+        if 'cpu_variant' in kwargs:
+            if 'debug' in kwargs['cpu_variant']:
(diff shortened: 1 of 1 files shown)


enjoy-digital (changes requested):
Thanks @kgugala, we should in fact probably use the new (optional) `add_soc_components` methods of CPUs for that: https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc.py#L899-L901.

You could add a `add_soc_components` method to VexRiscv CPU that adds the additional memory.

You can look at VexRiscv-SMP or Microwatt for examples:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/vexriscv_smp/core.py#L291-L319
https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/microwatt/core.py#L119-L130

The nice thing is that you'll only have to do it once and it will be available on all targets :)

kgugala:
Indeed this is cleaner. I opened https://github.com/enjoy-digital/litex/pull/774 (so I'll close this one)

PR #152, opened 2021-01-15
nexys4ddr: add etherbone support

gsomlo:
add etherbone (and `--with-etherbone` command line argument) to nexys4ddr.


enjoy-digital:
Thanks @gsomlo, merged.

Issue #148, opened 2021-01-07
ULX3S: Adding spiflash, U3 pinout error


blakesmith:
Not sure if I should open this issue here, or on some other ULX3S repo...

I'm attempting to add the onboard SPI flash to the ULX3S board to enable my SoC to boot from flash (See initial WIP commit here: https://github.com/blakesmith/litex-boards/commit/2859975144bbb5a4a872ca690dfb7907fbd8b8af#diff-0d05f744fae28b2e373b10e5bc6eda679ce414f074620c726c5044e1071ed72eR87). Relevant snippet is here:

```python
    # SPIFlash
    ("spiflash", 0,
        Subsignal("cs_n", Pins("R2")),
        Subsignal("clk", Pins("U3")),
        Subsignal("miso", Pins("V2")),
        Subsignal("mosi", Pins("W2")),
        Subsignal("wp", Pins("Y2")),
        Subsignal("hold", Pins("w1")),
        IOStandard("LVCMOS33")
    ),
    ("spiflash4x", 0,
        Subsignal("cs_n", Pins("R2")),
        Subsignal("clk", Pins("U3")),
        Subsignal("dq", Pins("W2", "V2", "Y2", "W1")),
        IOStandard("LVCMOS33")
    ),
```

I followed the [schematics here](https://github.com/emard/ulx3s/blob/master/doc/schematics_v314.pdf), but the "U3" pin is not present in the CABGA381 package, and get a constraint error:

```
Info: Packing IOs..                                                                     
Info: pin 'clk25$tr_io' constrained to Bel 'X0/Y47/PIOA'.                               
Info: pin 'oled_ctl_csn$tr_io' constrained to Bel 'X0/Y89/PIOA'.                                                                                                                
Info: pin 'oled_ctl_dc$tr_io' constrained to Bel 'X0/Y92/PIOA'.                                                                                                                 
Info: pin 'oled_ctl_resn$tr_io' constrained to Bel 'X0/Y92/PIOB'.                                                                                                               
Info: pin 'oled_spi_clk$tr_io' constrained to Bel 'X0/Y92/PIOD'.                        
Info: pin 'oled_spi_mosi$tr_io' constrained to Bel 'X0/Y92/PIOC'.                       
Info: pin 'rst$tr_io' constrained to Bel 'X4/Y95/PIOA'.                                 
Info: pin 'sdram_cas_n$tr_io' constrained to Bel 'X126/Y89/PIOA'.                       
[...]

PR #155, opened 2021-01-23
ULX3S: Make spiflash optionally accessible from the SoC, and bootable

blakesmith:
A follow on from: https://github.com/litex-hub/litex-boards/issues/148. I wanted to be able to use the onboard flash of the ULX3S as a boot target. This PR adds a new `--with-spiflash` flag that will add the SPI master to the SoC, and lift it into the memory map.

The easiest way to define the `FLASH_BOOT_ADDRESS` [pre-processor definition](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/bios/boot.c#L468) in the bios was to just do an `soc.add_constant` call. I couldn't find any other board examples defining `FLASH_BOOT_ADDRESS`, so if there's a better way to do this (or a cleaner pattern), please let me know!

Because [this commit](https://github.com/enjoy-digital/litex/commit/41964f945cee71337e8e0998d597086b2ef94f77) removed SPI flashing functionality from the bios, I built and tested a combined bitstream by using a modified version of @gregdavill's [combine script from the OrangeCrab ](https://github.com/gregdavill/OrangeCrab-examples/blob/main/litex/combine.py) to build a combined ECP5 bitstream and firmware image (thanks Greg!), which I was able to verify on my board.

Thanks!

### DIFF
--- a/litex_boards/platforms/ulx3s.py
+++ b/litex_boards/platforms/ulx3s.py
@@ -81,6 +81,21 @@ _io_common = [
         IOStandard("LVCMOS33")
     ),
 
+    # SPIFlash
+    ("spiflash", 0,
+        Subsignal("cs_n", Pins("R2")),
+        Subsignal("miso", Pins("V2")),
+        Subsignal("mosi", Pins("W2")),
+        Subsignal("wp", Pins("Y2")),
+        Subsignal("hold", Pins("W1")),
+        IOStandard("LVCMOS33")
+    ),
+    ("spiflash4x", 0,
+        Subsignal("cs_n", Pins("R2")),
+        Subsignal("dq", Pins("W2", "V2", "Y2", "W1")),
+        IOStandard("LVCMOS33")
+    ),
+
     # OLED
     ("oled_spi", 0,
         Subsignal("clk",  Pins("P4")),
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @blakesmith for the detailed PR and describing the reasoning behind it. This looks good, happy to merge it. I'll probably also apply this to the other boards with SPI Flash in the future.

blakesmith:
Thanks @enjoy-digital! Let me know if you want any help with any of the other boards. I don't have any other boards besides the ULX3S (not many others I could find that are still for sale!), but I'd be happy to help in any way I can, just let me know!

gregdavill:
It will probably be worth refactoring that combine script into a small LiteX utility at some point.

blakesmith:
@gregdavill I can take a crack at refactoring your `combine.py` script into something parameterized with command line arguments (was going to do that anyways), if you'd be willing to review. What interface do you think would be best to read in the `flash_regions`?

**JSON file with a flash region map**

Something like:

```json
{
  "flash_regions": [
    {
      "name": "bitstream",
      "file": "/some/file/path/bitsream.bit",
      "address": "0x80000000"
    },
    {
      "name": "firmware",
      "file": "/some/file/path/firmware.bin",
      "address": "0x80008000"
    }
  ]
}
```

Then:

```sh
$ flash_bitstream_combine.py --flash_regions regions.json -o combined.bit
```

**Vararg `argparse` parameters**

Something like:

```sh
$ flash_bitstream_combine.py --flash_region "bitstream: 0x80000000, /some/file/path/bitstream.bit", --flash_region "firmware: 0x80008000, /some/file/path/firmware.bin"  -o combined.bit
```

There's probably a better way to do this with python's argparse, but I'd have to dig in a bit more to see what's most natural.

I kind of like the JSON approach better, but what do you think?

PR #157, opened 2021-01-27
Add colorlight i5 board support

kazkojima:
Hi,

This is a proposed colorlight i5 board support. Only one thing that is different from other boards would be that the ethernet transceiver on this board requires to configure with mdio before using the ethernet. targets/colorlight_i5.py writes it to the newly created include/generated/target.h with

```
        os.makedirs(os.path.join(builder.software_dir, "include/generated"),
                    exist_ok=True)
        write_to_file(
            os.path.join(builder.software_dir, "include/generated", "target.h"),
            "// Colorlight i5 needs this to disable TX data to clock delay.\n"
	    "#define TARGET_BIOS_INIT_FUNC() mdio_write(0, 0x1c, 0x8c00)")
```

and also defines a macro TARGET_BIOS_INIT

```
            self.add_constant("TARGET_BIOS_INIT", 1)
```

when ethernet is enabled. I've added corresponding lines to litex software/bios/main.c which call TARGET_BIOS_INIT_FUNC() if TARGET_BIOS_INIT is defined.
I'm not sure that it's the right way to do. If looks not good, I'll be back with removing that part.

Regards,
	kaz

### DIFF
--- a/litex_boards/platforms/colorlight_i5.py
+++ b/litex_boards/platforms/colorlight_i5.py
@@ -1,0 +1,153 @@
+#
+# This file is part of LiteX-Boards.
+#
+# Copyright (c) 2021 Kazumoto Kojima <kkojima@rr.iij4u.or.jp>
+# SPDX-License-Identifier: BSD-2-Clause
+
+# The Colorlight i5 PCB and IOs have been documented by @wuxx
+# https://github.com/wuxx/Colorlight-FPGA-Projects
+
+from litex.build.generic_platform import *
+from litex.build.lattice import LatticePlatform
+from litex.build.lattice.programmer import EcpDapProgrammer
+
+# IOs ----------------------------------------------------------------------------------------------
+
+_io_v7_0 = [ # Documented by @smunaut
+    # Clk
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Hi @kazkojima, thanks for the board support (and for the nice https://github.com/kazkojima/colorlight-i5-tips repo).

For Ethernet, have you seen that it's also possible to adjust the TX/RX delay in the PHY? https://github.com/enjoy-digital/liteeth/blob/master/liteeth/phy/ecp5rgmii.py#L186-L187 Could you do a test with this? This could avoid having to do manual  adjustments in the BIOS.

Otherwise nice to see the recently added optional Wishbone interface on VexRiscv-SMP has also been useful to get Linux-on-LiteX-VexRiscv working on this board (that has DMs pins grounded).

kazkojima:
tx_delay=0 works magically! I've revised the pull-request.

Yes, the optional Wishbone interface solves the SDRAM issue on colorlight i5 and the linux can boot with VexRiscv SMP core.

enjoy-digital:
Great, thanks for the board support, this has been merged.

PR #163, opened 2021-01-30
Be friendlier about incompatible options.

garytwong:
Collect `--with-ethernet`/`--with-etherbone`, `--with-spi-sdcard`/`--with-sdcard`, etc. into `ArgumentParser.add_mutually_exclusive_group()`s.  That way, we get pretty `--help` output, and appropriate error messages if somebody tries to ask for something that doesn't make sense.


enjoy-digital:
Thanks @garytwong, I wasn't aware of `add_mutually_exclusive_group()`, interesting! This will be merged but I just want to have a closer look at it in the next days.

enjoy-digital:
This has been merged, thanks again @garytwong for this and for applying it to all boards.

garytwong:
Thanks very much!

PR #165, opened 2021-01-30
Add flash to SPI flash support for board ECPIX5

zguig52:
Needs also an update on openfpgaloader.py class
https://github.com/litex-hub/linux-on-litex-vexriscv/issues/189


enjoy-digital:
Thanks @zguig52, that's merged.

PR #167, opened 2021-02-01
sockit: Fix cable name, default to jtag_atlantic

hansfbaier:
Hi Florent,
I got a request from a user desperate for a serial connection
and I found this solution (include fix to programmer cable name).
Does lxterm work over jtag_atlantic?
(This I have not found out yet).
If not, what would I need to do to make it work?
Best regards,
Hans


enjoy-digital:
Hi @hansfbaier,

thanks, yes it should work, I'm not using it regularly but this hasn't changed since it was added. I can merge your PR.

Florent

PR #169, opened 2021-02-03
targets/colorlight_i5: use .bit stream instead of .svf when loading.

kazkojima:
Hi,

I've mistakenly thought ecpdap can load a .svf stream file, though it ends up with
```
Error: ECP5 status register in incorrect state.
```
after the file transfer is done. This PR fixes it.

Regards,
    kaz


enjoy-digital:
Thanks @kazkojima, that's merged.

kazkojima:
Thanks!

PR #172, opened 2021-02-08
sockit: Add an option to plug in an UART via the GPIO daughter board, make connector pin numbers one-based

hansfbaier:
jtag_atlantic is too slow to upload larger binaries.
We need a physical uart option.

### DIFF
--- a/litex_boards/platforms/arrow_sockit.py
+++ b/litex_boards/platforms/arrow_sockit.py
@@ -147,26 +147,33 @@ _io = [
         Subsignal("aud_i2c_sdat", Pins("AF30")),
         Subsignal("aud_mute",     Pins("AD26")),
         IOStandard("3.3-V LVTTL")
-    )
+    ),
+
+    ("gpio_serial", 0,
+        Subsignal("tx", Pins("J3:9")),
+        Subsignal("rx", Pins("J3:10")),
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @hansfbaier, it could be done more easily with
- Adding `gpio_serial` pins to the platform.
- Just using `./target.py --uart-name=gpio_serial` to select the `gpio_serial ` instead of JTAG?

Would you mind doing the changes on this?

enjoy-digital (changes requested):
Thanks @hansfbaier, it could be done more easily with
- Adding `gpio_serial` pins to the platform.
- Just using `./target.py --uart-name=gpio_serial` to select the `gpio_serial ` instead of JTAG?

Would you mind doing the changes on this?

hansfbaier:
@enjoy-digital OK, I made the changes and rebased my branch to newest upstream.
I was not aware that add_uart automatically requests the uart name pins, and this is more elegant, yes.
Although not as explicit for the user.
(He will have to look at the source code to find out which uart name to choose)

hansfbaier:
@enjoy-digital The requested changes are ready for review.

enjoy-digital (approved):
Thanks @hansfbaier, this looks good.

PR #174, opened 2021-02-09
vc707.py: clk156 add missing constraint

yetifrisstlama:
The `create_clock -name sys_clk` entry was missing in the generated .xdc file, which leads to critical warnings in synthesis.

Adding this period constraint fixes it.


enjoy-digital:
Thanks @yetifrisstlama, merged.

PR #176, opened 2021-02-16
sds1104xe: add offsetdac

tmbinc:
Add offsetdac&offsetmux, which are used on SDS1104X-E to set the offset voltage for each of the 4 channels with a single high-precision DAC and an analog MUX.

The state machine continuously refreshes all 8 inputs by setting the DAC, setting the Mux, and then relying on the external analog sample&hold logic to keep the value. To prevent drift, this is done with a relatively high rate.

### DIFF
--- a/litex_boards/platforms/sds1104xe.py
+++ b/litex_boards/platforms/sds1104xe.py
@@ -66,36 +66,50 @@ _io = [ # Documented by https://github.com/360nosc0pe project.
         IOStandard("LVCMOS33"),
     ),
 
+    # Scope things
+    ("offsetmux", 0,
+        Subsignal("S",       Pins("U14 Y18 AA18")),
+        Subsignal("nE",      Pins("U19")),
+        IOStandard("LVCMOS15"),
+    ),
+
+    ("offsetdac", 0,
+        Subsignal("SCLK",       Pins("H15")),
+        Subsignal("DIN",        Pins("R15")),
+        Subsignal("nSYNC",      Pins("J15")),
+        IOStandard("LVCMOS15"),
+    ),
+
     # DDR3 SDRAM
     ("ddram", 0,
         Subsignal("a", Pins(
             "J21 K18 J18 R16 P16 T18 R18 T19",
             "R19 P18 P17 P15 N15"),
-            IOStandard("SSTL135")),
-        Subsignal("ba",    Pins("K21 J20 J22"), IOStandard("SSTL135")),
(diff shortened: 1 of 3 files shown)


litex_boards/platforms/sds1104xe.py
```diff
@@ -66,36 +66,50 @@
...
+        Subsignal("nSYNC",      Pins("J15")),
+        IOStandard("LVCMOS15"),
+    ),
+
     # DDR3 SDRAM
     ("ddram", 0,
         Subsignal("a", Pins(
             "J21 K18 J18 R16 P16 T18 R18 T19",
             "R19 P18 P17 P15 N15"),
-            IOStandard("SSTL135")),
```
tmbinc, line 74: Some/all of the DDR pins are shared with LVCMOS15 pins. We need to double check what the IO standard for the DDR really is.

litex_boards/targets/sds1104xe.py
```diff
@@ -36,6 +36,8 @@
 
 from liteeth.phy.mii import LiteEthPHYMII
 
+from sds1104xe_offsetdac import OffsetDac
```
tmbinc, line 39: Where do we put more complex per-board logic? This component can likely be shared on similar hardware (such as the SDS1202X-E, the 2-channel variant).

litex_boards/targets/sds1104xe_offsetdac.py
```diff
@@ -0,0 +1,164 @@
...
+    def __init__(self, offsetdac_pads, offsetmux_pads):
+        self._status = csr.CSRStatus(32, name='status', reset=0x0ffdac)
+        self.spi = Record(SPI)
+        self.mux = Record(MUX)
+
+        self.comb += self.spi.connect(offsetdac_pads)
+        self.comb += self.mux.connect(offsetmux_pads)
+
+        self._control = csr.CSRStorage(32, name='control') # DAC PD control
+        self._clkdiv = csr.CSRStorage(32, name='clkdiv', reset=326) # roughly matches original timing
```
tmbinc, line 25: This needs to be fixed corresponding to the sysclk

enjoy-digital:
Thanks @tmbinc, as suggested on IRC, LiteX-Boards is maybe not a right place for custom peripherals, maybe we could create a project under https://github.com/360nosc0pe for the specific scope part? And just try to base it on LiteX-Boards to avoid duplicating base component support (DRAM, Ethernet, etc...).

tmbinc:
That makes sense - we'll maintain this under 360nosc0pe for now.

PR #177, opened 2021-02-18
target/arty: add eth_ip_configurable switch

AleksandraSwierkowska:
This is a companion PR to enjoy-digital/litex#822

This PR adds eth_ip_configurable switch to the arty.py target. The switch, used with the ethernet switch, allows to change local IP, remote IP and MAC address in runtime. It should not be used with the etherbone switch as it has no affect on the etherbone address.

### DIFF
--- a/litex_boards/targets/arty.py
+++ b/litex_boards/targets/arty.py
@@ -56,7 +56,7 @@ class _CRG(Module):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCCore):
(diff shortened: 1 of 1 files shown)


AleksandraSwierkowska:
Changed eth_ip_configurable into eth_dynamic_ip to match the naming convention in the litex/soc/software/bios

enjoy-digital:
Thanks @AleksandraSwierkowska for the PR and changes/update. This is merged.

PR #179, opened 2021-02-24
Numato sdcard pmod for Arty

shenki:


dayjaby:
Looking good! I tested the same pmod description and was able to write to/read from the SD card.

enjoy-digital:
Thanks @shenki, this is merged.

PR #183, opened 2021-03-05
crosslink_nx_vip: Camera IO fixes

gatecat:
While testing I want to feed a MCLK to the camera so I can test I2C communications, without creating unused MIPI pins that cause a failure because the hard IP isn't instantiated yet. In general, it is likely that MCLK and the MIPI pins are likely to be used in different parts of the design anyway.

Also, fix the configuration of `cam_reset`.


enjoy-digital:
Thank @gatecat, that's merged.

PR #185, opened 2021-03-07
arty: Add an option to enable jtagbone

stffrdhrn:
Then adds jtagbone for arty.  I have tested with the following
litex_server and it seems to work fine.

  litex_server --jtag --jtag-config openocd_xc7_ft2232.cfg

Note, the jtagbone and etherbone may be mutually exclusive, but I am not
sure how to define that in the args.


stffrdhrn:
Ill update my branch, there is a conflict

enjoy-digital:
Thanks @stffrdhrn, JTAGBone and Etherbone are not mutually exclusive so this is fine like this. If this can be useful, mutually exclusive args can be created like this: https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/nexys_video.py#L156-L158

stffrdhrn:
> Thanks @stffrdhrn, JTAGBone and Etherbone are not mutually exclusive so this is fine like this. If this can be useful, mutually exclusive args can be created like this...

I think the complexity I was having was that --etherbone is already mutually exclusive with --ethernet. So pairing --etherbone with-- jtagbone too might need a different trick. 
But it's good it's not a problem here.

PR #186, opened 2021-03-08
crosslink_nx_vip: Remove constraints for hard MIPI pins

gatecat:
Working around the Radiant bug that affects these pins


enjoy-digital:
Thanks @gatecat, that's merged.

PR #187, opened 2021-03-16
arrow_sockit: get video terminal working on VGA

hansfbaier:
litex-sockit
It's not perfect yet (the color timing of individual components seems to be off by one pixel,
but it is definitely useable.

### DIFF
--- a/litex_boards/platforms/arrow_sockit.py
+++ b/litex_boards/platforms/arrow_sockit.py
@@ -110,10 +110,11 @@ _io = [
 
     # VGA
     ("vga", 0,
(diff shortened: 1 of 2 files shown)


hansfbaier:
It looks like the red channel is exactly one pixel late.

hansfbaier:
I just updated the resolution to 1024x768@60. Looks much nicer and works too.

enjoy-digital:
Nice, thanks for the test and PR. As you see there are still a few things to fix (line clear for example), but  this can already be convenient for some use cases.

PR #188, opened 2021-03-18
fix #848: allow ram initialization in bitstream to enable block ram 

hansfbaier:
This is a fix for #848 [https://github.com/enjoy-digital/litex/issues/848](https://github.com/enjoy-digital/litex/issues/848)


enjoy-digital:
Thanks @hansfbaier, that's something interesting to know about Max10 devices.

PR #191, opened 2021-03-23
nexys_video: choose VADJ value with arguments

acomodi:
Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>

The VADJ value changes the resistance to set the voltages for the FMC expansion board. Therefore, depending on the FMC expansion board we need to be able to modify the VADJ value when building the nexys_video target.

This PR adds the `--vadj` argument to choose this value.

### DIFF
--- a/litex_boards/targets/digilent_nexys_video.py
+++ b/litex_boards/targets/digilent_nexys_video.py
@@ -60,7 +60,7 @@ class _CRG(Module):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCCore):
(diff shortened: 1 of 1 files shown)


enjoy-digital (changes requested):
Thanks @acomodi, we could indeed merge this, but I think it would be better to directly allow user to specific the voltage and  let the code translate this directly to the binary values:

image

Could you update the PR with this? (ex: --vadj=1.2V, 1.8V 2.5V 3.3V).
I also did some renaming on platforms (with retro-compatibly on imports), can you also update the PR to use the new naming?

Thanks

acomodi:
@enjoy-digital Sorry for the delay. I pushed a fix to have textual VADJ arguments and rebased the branch on top of master.

enjoy-digital:
Thanks @acomodi for the changes and rebase.

PR #193, opened 2021-03-27
ztex213 fix; propagate variant from targets to platform

rdolbeau:
Otherwise the variant is ignored...


enjoy-digital:
Thanks @rdolbeau, that's merged.

PR #199, opened 2021-04-10
boards/compat: Skip targets with unsatisfied dependencies.

### LINKED ISSUE #194: project requires installing dependencies for all platforms/targets
I've recently updated the migen/litex dependencies in a project tree and saw that litex-boards had been split off into it's own repo (yes, I'm slow on the update .. :))

Looking at https://github.com/litex-hub/litex-boards/blob/master/litex_boards/__init__.py , it seems like now litex will try to install dependencies for every possible platform+target, even if it's not actually being used. Or am I just doing something incorrectly? For now I've just locally modified `__init__.py` to reference only the single target I care about (lattice_versa_ecp5).

zyp:
Fixes #194

### DIFF
--- a/litex_boards/__init__.py
+++ b/litex_boards/__init__.py
@@ -58,7 +58,11 @@ for target in targets:
     # from litex_boards.targets import digilent_arty or
     # from litex_boards.targets import arty
     if target.split("_")[0] in vendors:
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @zyp for looking at this, that's merged.

PR #200, opened 2021-04-10
Qmtech Wukong: updates

rdolbeau:
fix ethernet clock (it's a GMII), add FB support over the HDMI connector (hdmi clock set from the resolution)

This and https://github.com/enjoy-digital/litex/pull/875 enable a working FB in Linux-on-Litex-Vexriscv.

This and forcing polling mode in Linux (see https://github.com/enjoy-digital/liteeth/issues/62 ) enable a working Ethernet in Linux-on-Litex-Vexriscv.

### DIFF
--- a/litex_boards/platforms/qmtech_wukong.py
+++ b/litex_boards/platforms/qmtech_wukong.py
@@ -77,10 +77,10 @@ _io = [
         Misc("SLEW=FAST"),
     ),
 
-    # MII Ethernet
-    ("eth_ref_clk", 0, Pins("U1"), IOStandard("LVCMOS33")),
+    # GMII Ethernet
     ("eth_clocks", 0,
         Subsignal("tx", Pins("M2")),
+        Subsignal("gtx", Pins("U1")),
         Subsignal("rx", Pins("P4")),
         IOStandard("LVCMOS33")
     ),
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @rdolbeau, that's merged.

mdejw:
The board has GMII PHY, but there is defined MII PHY. Why?
This does not work good on my side - transmission is not stable.

I've prepared PR:
https://github.com/litex-hub/litex-boards/pull/221

rdolbeau:
@mdejw It's not defined as MII; LiteEthPHY correctly auto-detect that the Phy is a GMII/MII 10/100/1000 (https://github.com/enjoy-digital/liteeth/blob/435c67dbc774ab2f49c6d53f5f63417ba4af7605/liteeth/phy/__init__.py#L6-L14).

This works fine to establish a link in both Gigabit and FastEthernet (depending on what the other side is using); However in Gigabit mode, the core is likely too slow to keep up with the link in Linux - you get lots of dropped packet. FastEthernet has no issue.

What are the symptoms you're trying to fix?

mdejw:
I had a problem with loading image via TFTP on Litex BIOS. I've tried only on 1GBit link speed.
I've seen on Wireshark only one ARP request from Litex board and ARP reply from server.

Thank you, I hadn't look that deep into implementation.

I assumed it was an MII PHY as I saw following lines on Litex netboot:
```
Ethernet phy mode: MII

```
So - seems it does not detect GMII/1GBit on my side.

rdolbeau:
@mdejw The GMII/MII implementation default to MII rather then GMII if the system clock is below 125 MHz (https://github.com/enjoy-digital/liteeth/blob/435c67dbc774ab2f49c6d53f5f63417ba4af7605/liteeth/phy/gmii_mii.py#L115-L119). If you boot Linux then the driver will use gigabit.

If it doesn't work for you, perhaps you can try forcing the mode (e.g. `mode = modes["GMII"]` there: https://github.com/enjoy-digital/liteeth/blob/435c67dbc774ab2f49c6d53f5f63417ba4af7605/liteeth/phy/gmii_mii.py#L183) to see if it helps. Or try with a FastEthernet switch if you have one lying around.

PR #202, opened 2021-04-22
terasic_deca: Add MiSTer SDRAM module support

hansfbaier:
Hello,
I implemented this on request of @somhi
He tested it and confirmed it works successfully:
https://github.com/SoCFPGA-learning/DECA/tree/main/Projects/sdram_mister_deca

### DIFF
--- a/litex_boards/platforms/terasic_deca.py
+++ b/litex_boards/platforms/terasic_deca.py
@@ -4,6 +4,10 @@
 # Copyright (c) 2014-2019 Hans Baier <hansfbaier@gmail.com>
 # SPDX-License-Identifier: BSD-2-Clause
 
+# To use the MiSTer SDRAM option, please connect the SDRAM
+# module as described here:
+# https://github.com/SoCFPGA-learning/DECA/tree/main/Projects/sdram_mister_deca
+
 from litex.build.generic_platform import *
 from litex.build.altera import AlteraPlatform
 from litex.build.altera.programmer import USBBlaster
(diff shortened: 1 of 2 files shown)


hansfbaier:
image

hansfbaier:
I actually might start to be inclined to move the SDRAM chip to litedram at last.
Many of the QMTech boards have the same chip and while this only runs well at certain frequencies,
it still might be useful if the comments make it clear where it operates well.

enjoy-digital:
Hi @hansfbaier,

that's nice to see you got it working. This combination (Terasic Deca + MiSTer SDRAM) seems very specific to this project. Since the instructions are already in an external repository, I think it would make more sense to move the specific target for this use case also to this repository, this would also simplify maintenance on my side. In LiteX-Boards, we could just have a link to this external repository. What do you think @hansfbaier, @somhi?

hansfbaier:
@enjoy-digital  Yes, that makes sense, good idea. @somhi: What do you think?

somhi:
If for you is Ok, then it's also Ok for me! 🙂

enjoy-digital:
@hansfbaier, @somhi, perfect thanks. BTW, using `platform.add_extension` you can define the SDRAM pins directly in the target file, similarly to what is done here to define the SDCard IOs: https://github.com/enjoy-digital/litesdcard/blob/master/bench/trellisboard.py#L40-L50. This way for this configuration, you can rely on the platform from LiteX-Boards and just need a specific target.

hansfbaier:
OK, code moved to @somhi 's repository. Pull request: https://github.com/SoCFPGA-learning/DECA/pull/1

somhi:
I've tested new Deca target now in external repo and is building Ok.

@enjoy-digital you said "In LiteX-Boards, we could just have a link to this external repository.".
Should I add it and where, or you do it?

Thanks

enjoy-digital:
@somhi: Sure, we can add it. Can you share the link here? I'll add it.

somhi:
The link to the alternative Deca target is https://github.com/SoCFPGA-learning/DECA/raw/main/deca_litex_target.py
The link to the project is https://github.com/SoCFPGA-learning/DECA/tree/main/Projects/sdram_mister_deca
At the end of the project page there are instructions to download and build the alternative Deca target.

somhi:
By the way, I'm trying to do the Linux on Litex with Deca and the Mister SDRam module.  It builds Ok and I upload image Ok, but it hangs after Liftoff.  Last time I spoke with @hansfbaier he had the same issue with Arrow Sockit and Mister SDRam module.
I've also replaced rv32.dtb with the one compiled from the dts file that is generated in the build folder.

PR #203, opened 2021-04-25
arty: add option to use litespi QSPI controller

kgugala:
Signed-off-by: Karol Gugala <kgugala@antmicro.com>

### DIFF
--- a/litex_boards/targets/digilent_arty.py
+++ b/litex_boards/targets/digilent_arty.py
@@ -34,13 +34,16 @@ from litespi import LiteSPI
 # CRG ----------------------------------------------------------------------------------------------
 
 class _CRG(Module):
-    def __init__(self, platform, sys_clk_freq):
+    def __init__(self, platform, sys_clk_freq, with_mapped_flash):
         self.rst = Signal()
         self.clock_domains.cd_sys       = ClockDomain()
         self.clock_domains.cd_sys4x     = ClockDomain(reset_less=True)
         self.clock_domains.cd_sys4x_dqs = ClockDomain(reset_less=True)
         self.clock_domains.cd_idelay    = ClockDomain()
         self.clock_domains.cd_eth       = ClockDomain()
+
+        if with_mapped_flash:
+            self.clock_domains.cd_qspi  = ClockDomain() # we need a domain with reset for litespi
 
         # # #
 
@@ -52,6 +55,10 @@ class _CRG(Module):
         pll.create_clkout(self.cd_sys4x_dqs, 4*sys_clk_freq, phase=90)
         pll.create_clkout(self.cd_idelay,    200e6)
         pll.create_clkout(self.cd_eth,       25e6)
+
+        if with_mapped_flash:
+            pll.create_clkout(self.cd_qspi,      4*sys_clk_freq)
+
         platform.add_false_path_constraints(self.cd_sys.clk, pll.clkin) # Ignore sys_clk to pll.clkin path created by SoC's rst.
 
         self.submodules.idelayctrl = S7IDELAYCTRL(self.cd_idelay)
@@ -71,7 +78,7 @@ class BaseSoC(SoCCore):
             **kwargs)
 
         # CRG --------------------------------------------------------------------------------------
-        self.submodules.crg = _CRG(platform, sys_clk_freq)
+        self.submodules.crg = _CRG(platform, sys_clk_freq, with_mapped_flash)
 
         # DDR3 SDRAM -------------------------------------------------------------------------------
         if not self.integrated_main_ram_size:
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @kgugala, we can merge it. Can you just give a quick status of LiteSPI with QSPI Flash? Is it working fine / fast? (I remember I wanted to use it on Fomu last year, got it working IIRC but it was a lot slower than the current LiteX SPI Flash core. Ideally we should switch to LiteSPI core, since a lot cleaner/easier to understand, but we should also make sure resource usage is minimal and speed is acceptable, so I'm just wondering if you have some feedback on this? I could also try to do some comparisons between both cores on resources/speed to understand things that can be improved in LiteSPI.

kgugala:
The problem with FOMU was that it was really hard to fit there CDC logic, preventing us from clocking the flash faster than the sys_clk. I'll update this PR to use different clocks.

litex_boards/targets/digilent_arty.py
```diff
@@ -93,6 +99,18 @@ def __init__(self, variant="a7-35", toolchain="vivado", sys_clk_freq=int(100e6),
         if with_jtagbone:
             self.add_jtagbone()
 
+        if with_mapped_flash:
+
+            self.submodules.spiflash_phy  = LiteSPIPHY(platform.request("spiflash"), S25FL128S(Codes.READ_1_1_1))
+            self.submodules.spiflash_mmap  = LiteSPI(phy=self.spiflash_phy,
+                clk_freq        = sys_clk_freq,
+                mmap_endianness = self.cpu.endianness)
+            self.add_csr("spiflash_mmap")
```
enjoy-digital, line 108: self.add_csr is no longer required (still works but now automatically collected/assigned).

litex_boards/targets/digilent_arty.py
```diff
@@ -118,6 +136,7 @@ def main():
     parser.add_argument("--sdcard-adapter",      type=str,                         help="SDCard PMOD adapter: digilent (default) or numato")
     parser.add_argument("--no-ident-version",    action="store_false",             help="Disable build time output")
     parser.add_argument("--with-jtagbone",    action="store_true",              help="Enable Jtagbone support")
+    parser.add_argument("--with-mapped-flash", action="store_true",             help="Enabel memory mapped flash")
```
enjoy-digital, line 139: Enable.

litex_boards/targets/digilent_arty.py
kgugala: removed

litex_boards/targets/digilent_arty.py
kgugala: fixed

kgugala:
@enjoy-digital I have moved the QSPI PHY to a separate 4x faster clock domain.

enjoy-digital:
@kgugala OK, I'm going to merge it. Using a 400MHz clock internally seems a bit strange, if this is really required, that's possible there are some bottlenecks in the core. With this merged, I'll also play with it and will try to compare the SPI core we are actually using on the other targets. We can experiment on Arty and could use this core on the other targets if we manage to get something closer in performance and resource to the actual core.

enjoy-digital:
(that's also possible we'll change the argument name in the future).

kgugala:
thanks.

the 400MHz clock is for the PHY part only (core still runs at sys_clk). This is to fetch the data from the chip faster.

mithro:
@kgugala - I'm pretty sure the SPI flash can't run at 400 MHz?

enjoy-digital:
@kgugala, @mithro: I've been able to test it and did some simplifications/cleanups:
- https://github.com/litex-hub/litex-boards/commit/5bfeb999e4c0766b93f06d2a8b4c74ceb4956767
- https://github.com/litex-hub/litespi/commit/c9934abed8ce8558e2e15aeb9d3f56ae595c52d0
- https://github.com/litex-hub/litespi/commit/9f02ebbe6bd12328345b6fdedc2b20654d071898

Timings were failing at 400MHz, so I reverted to sys_clk, I don't think a higher clock is required on Arty (or there would probably be an architecture issue in the core :)). With sys_clk, reads seems to be working fine and I got ~5MB/s speed from the CPU in Quad mode.

If we want to extend the use of LiteSPI, that could be useful to have someone working on it to:
- Compare resource usage/performance against the actual core: https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/spi_flash.py. (Also do performance measurement  in XiP).
- Minimize resources if difference is significant, improve performance if difference is significant.
- Be sure we can write to the Flash in Bit-Banging mode.

mithro:
I agree that LiteSPI version should use equal or less resources than the existing core.

enjoy-digital:
@kgugala @mithro: I've been working on moving the targets to LiteSPI and have been able to do some comparisons of resource usage: https://github.com/litex-hub/litex-boards/commit/0f648ac4efbf65ee0ec72d51d52dcb0d365fb568

Most of the additional resources seems to come from the Master module, when disabling it, we already get better resource usage:
https://github.com/litex-hub/litex-boards/commit/12fb315e094f1b33d55769e9213de8ce8db8037c

Most of this seems to come from the internal FIFO, by exposing the fifo_depth parameter to allow user to reduce it on small FPGAs we already have interesting results: 
https://github.com/litex-hub/litespi/commit/8e90a1d05ed1e9414ea3ccff8499a846a577db6f
https://github.com/litex-hub/litex-boards/commit/533d25e845227e6fe3f4a3c550d5b2d97f127d86

We could probably optimize this further, but this is already good enough to start using it and all the targets that were using the old SPI Flash core are now switched to LiteSPI (and tested).

@kgugala: Just for info, on Arty I renamed --with-mmaped-flash to --with-spi-flash: https://github.com/litex-hub/litex-boards/commit/4e2b596ab3ad114e7b6bc4ae1104e5435790d7fa so you'll probably have to update the command you were using.

mithro:
@kgugala - @enjoy-digital work is here very interesting.

PR #204, opened 2021-04-27
terasic_sockit: fix: make video clock also optional as video terminal is optional

hansfbaier:
The video clock also needs to go away if the video terminal is not instantiated,
otherwise the project will have a build error.


enjoy-digital:
Thanks @hansfbaier, this is merged.

PR #205, opened 2021-04-27
antmicro_lpddr4_test_board: fix ethernet rx delay issue

jedrzejboczar:
We were using wrong RX delay which caused random errors as the PHY chip already introduces 1.2 ns delay.


enjoy-digital:
Thanks @jedrzejboczar, this is merged.

PR #209, opened 2021-04-29
de0nano-target: expose platform to inheriting classes for easier customization

hansfbaier:
This is necessary for inheriting targets to access the platform (eg to add extensions).


enjoy-digital:
Thanks @hansfbaier, this is in fact already addressed in LiteX directly https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc.py#L723 so is not required.

PR #211, opened 2021-05-04
Lattice iCE40 UltraPlus Breakout board (iCE40UP5K-B-EVN) added

Acathla-fr:
Platform from migen, modified a little bit. Target inspired by FOMU (same FPGA but without USB).
Needs an external serial connection.


enjoy-digital:
Thanks @Acathla-fr, this is merged.

PR #212, opened 2021-05-05
add QMTECH XC7A35T core board + daughter board

hansfbaier:
I tested litex running on the core board. DDR memory works fine.
I tested the daughterboard (leds + buttons) with my nmigen-boards version of this
(see https://github.com/nmigen/nmigen-boards/pull/152)

Unfortunately my core board now is wired up to a custom project,
and it would be time consuming to tear it apart.

It builds with Vivado and I feel it is definitely worth sharing,
especially to avoid duplication of effort.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,39 +119,41 @@ The Colorlight5A is a very nice board to start with, cheap, powerful, easy to us
 [> Development boards
 ---------------------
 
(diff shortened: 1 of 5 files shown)


enjoy-digital:
Thanks @hansfbaier, this is merged. The QMTECH boards are indeed nice cheap development boards and the daughter board support will be interesting for the other QMTECH boards (ex ecp4e15).

PR #214, opened 2021-05-08
icesugar: map the usb port

hansfbaier:
I overlooked the FPGA connected USB port here


enjoy-digital:
Thank @hansfbaier, that's merged.

PR #215, opened 2021-05-08
Qmtech xc7a35t

hansfbaier:
Here are two small fixes for the qmtech board.


enjoy-digital:

PR #217, opened 2021-05-15
QMTech EP4CE15: Add daughterboard support, small DECA addition

hansfbaier:
First application of the daughterboard outside the XC7A35T.


enjoy-digital:
Thanks @hansfbaier, this looks good. I'm just not sure to understand the reason of the added Video Framebuffer on the Terasic Deca? Since we don't have  DDR3 support on it now, the framebuffer will not work. If this is an error, can you remove this commit from the PR? (if that's too complicated, I can do it).

hansfbaier:
OK done, was not difficult at all :)

enjoy-digital:
Thanks for the update @hansfbaier, this is merged.

PR #221, opened 2021-05-25
Fix for Ethernet PHY type for Qmtech Wukong board

mdejw:
Fix for Ethernet PHY type for Qmtech Wukong board


rdolbeau:
See https://github.com/litex-hub/litex-boards/pull/200#issuecomment-850972380

enjoy-digital:
As pointed by @rdolbeau, the current PHY is probably correct so I'm closing. @mdejw in case you still have the issue please re-open.

PR #223, opened 2021-06-06
Fix colorlight 5a 75b v6.1 flash pins

david-sawatzke:
As with the other versions, the clk pin can't be directly driven and has to be used with the usrmclk primitive. Tested, reading works


enjoy-digital:
Thanks, that's merged.

PR #228, opened 2021-06-17
target/digilent_arty.py: Add with-save-env switch

MateuszKosmala:
That Pull Request adds with-save-env switch to digilent_arty.py target.
The switch allows to save and print environment variables to persistent storage and must be used together with eth_dynamic_ip switch what's assert in BaseSoC class.
That PR is associated with enjoy-digital/litex#947

I want to emphasize that memory alignment of spiflash in predefined address(code below) is not an optimal solution and proposed generic fix is associated with issue:  enjoy-digital/litex#959

```
mem_map =  {         
          "spiflash": 0x01000000,      
}
mem_map.update(SoCCore.mem_map)
```

### DIFF
--- a/litex_boards/targets/digilent_arty.py
+++ b/litex_boards/targets/digilent_arty.py
@@ -61,7 +61,13 @@ class _CRG(Module):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCCore):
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Closing, the feature would need to be discussed first.

Issue #235, opened 2021-06-29
Thoughts on adding the option to remove the LedChaser submodule from targets?


JosephBushagour:
The LedChaser submodule is a great diagnostic tool but it is somewhat costly in terms of LCs (around 70 LCs on ice40up5k boards like Fomu/iCEbreaker). I was thinking of submitting a PR that adds a `led_chaser=True` keyword argument to BaseSoCs of ice40up5k boards so that a user can turn off the LedChaser if they don't want it, but first I wanted to hear your thoughts on if that's a good or bad idea.

enjoy-digital:
@JosephBushagour That's a good idea yes, maybe we could just name it `with_led_chaser` for consistency with the other optional features (`with_pcie`, `with_etherne`t, etc...).

smosanu:
I'd also prefer having `with_led_chaser` as an option because my ES1 board has an issue with the way the LEDs are soldered and one of the LEDs creates a short that crashes the entire machine, so I had to disable it so far anyway.

enjoy-digital:
Thanks for PR #236 @JosephBushagour, this has been merged so we can close this.

PR #236, opened 2021-07-07
Add with_led_chaser argument to constructor of boards using LedChaser submodule.

JosephBushagour:
Related Issue: #235 

For consistency's sake, I've added the `with_led_chaser=True` kwarg to all boards that use the `LedChaser`, but I'd be happy to only apply the change to some boards if you'd rather do that.


tcal-x (approved):
LGTM

JosephBushagour:
@enjoy-digital Are you still feeling good about this change? If it's too large I could specifically target it to ice40up5k boards :).

enjoy-digital:
@JosephBushagour: Sorry for the delay, this is perfect, thanks for applying it to all boards!

PR #240, opened 2021-07-16
Alveo U280 peripherals and HBM2

smosanu:
The au280 platform file now has pins declared for PCIe and some of the other peripherals as well as an XDC entry that leads to synthesis completion for the HBM IP. The au280 target file allows building a system either with DRAM or with HBM. In the next steps I will make it possible to have both DRAM and HBM in a single system. Also, for the system with only HBM, I will attempt to develop an "add_hbm" method (similar to "add_sdram") to be able to boot linux just on HBM.


enjoy-digital:
Thanks @smosanu, this has been integrated manually with https://github.com/litex-hub/litex-boards/commit/99ff82c75a4bd0a4dfff1d0689f258846c21373e. I avoided putting the .xci directly in the repository for now (It's present as a github attachement and downloaded during the build). We'll improve this in the future.

PR #243, opened 2021-07-18
add FPGA Boards (Digilent CMOD A7, Digilent Nexys 4, Micronova Mercury2)

tweakoz:
1.  Digilent CMOD A7
      https://reference.digilentinc.com/programmable-logic/cmod-a7/start

2. Digilent Nexys 4
      https://reference.digilentinc.com/programmable-logic/nexys-4/start

3. MicroNova Mercury 2
      https://www.micro-nova.com/mercury-2

### DIFF
--- a/litex_boards/platforms/digilent_cmod_a7.py
+++ b/litex_boards/platforms/digilent_cmod_a7.py
@@ -1,0 +1,80 @@
+#
+# This file is part of LiteX-Boards.
+#
+# Copyright (c) 2018 William D. Jones <thor0505@comcast.net>
+# Copyright (c) 2020 Staf Verhaegen <staf@fibraservi.eu>
+# SPDX-License-Identifier: BSD-2-Clause
+
+from litex.build.generic_platform import Pins, Subsignal, IOStandard, Misc
+from litex.build.xilinx import XilinxPlatform
+from litex.build.openocd import OpenOCD
+
+# IOs ----------------------------------------------------------------------------------------------
+
+_io = [
+    # Clk / Rst
+    ("clk12", 0, Pins("L17"), IOStandard("LVCMOS33")),
+
+    ## Buttons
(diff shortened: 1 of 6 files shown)


enjoy-digital:
Thanks @tweakoz, this looks fine and we should be able to merge. I think you created the `AsyncRAM`/`CellularRAM `modules but don't see your copyright in the files, do you explicitly want to avoid having your copyright or is it missing?

tweakoz:
> Thanks @tweakoz, this looks fine and we should be able to merge. I think you created the `AsyncRAM`/`CellularRAM `modules but don't see your copyright in the files, do you explicitly want to avoid having your copyright or is it missing?

Honestly, did not think much about it  - I did create them - I was just happy to get it work.  I can add a copyright - another question - should those modules be part of a shared ip library or just leave them inline in the board target file? And I noticed I should also fix a few copy paste items (I think the CMod ident string states it is an Arty because I copied and pasted Arty as a starting point to bring Cmod up) - I will post an update this weekend.

enjoy-digital:
Hi @tweakoz,

with the copyright question, I was also anticipating some changes in the code: Moving the new `AsyncRAM`/`CellularRAM` modules to `litex/soc/cores/ram`, this would simplify reuse and later extend them with parameters. I could do these changes directly and with your copyright on the files, I'll be able to conserve these while moving the code. So if you can just update the targets/platforms with the copyrights, I could merge it and do changes on top of it.

tweakoz:
In that case,Here is my copyright line:
Copyright (c) 2020 Michael T. Mayers <michael@tweakoz.com>

enjoy-digital:
Thanks @tweakoz, the PR has been merged and the copyright added. We'll probably move the AsyncRAM/CellularRAM modules to LiteX directly in the future to simplify reuse.

PR #248, opened 2021-07-28
Add option for different Fomu SPI ICs.

JosephBushagour:
I've been very excited to see the great work on LiteSPI integration!

Fomu tools support 4 different SPI chips depending on the revision/type of your board (see https://github.com/im-tomu/fomu-flash/blob/c2576a62065b94dc2a47fb7820c8834937ba0343/spi.c#L632-L670). This PR adds support for all of the different ICs based on a keyword argument, while keeping the default as `AT25SF161`. One note is that the `GD25Q16C` currently doesn't support `READ_1_1_4` in LiteSPI.


enjoy-digital:
Thanks @JosephBushagour, this is merged. I just did minor changes with https://github.com/litex-hub/litex-boards/commit/54cee0598677861d9694438b7c7eab168eb8ce0f.

PR #251, opened 2021-08-09
FIX: OrangeCrab Feather SPI pad name

niw:
**Problems**

`SPIMaster` pad names are `clk`, `cs_n`, `mosi`, and `miso`.
However, `feather_spi` is using `sck` instead of `clk`, therefore
it is not able to use as-is for `SPIMaster`, for example,
with `add_spi` on Linux On LiteX VexRiscv.

**Solution**

In fact, `spisdcard` and other SPI related pad names are
using `clk`, only `feather_spi` is using `sck`.
Therefore, rename `sck` to `clk`.

### DIFF
--- a/litex_boards/platforms/gsd_orangecrab.py
+++ b/litex_boards/platforms/gsd_orangecrab.py
@@ -207,10 +207,10 @@ feather_i2c = [
 ]
 
 feather_spi = [
-    ("spi",0,
+    ("spi", 0,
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, sorry for the delay.

niw:
Not at all, thank you so much for all LiteX and related projects!

PR #257, opened 2021-08-16
terasic DECA: set IO bank voltages to prevent damage by accidental misconfiguration

hansfbaier:
This is necessary for preventing the user damaging the board unawares, I think.


enjoy-digital:
Thanks @hansfbaier, we are trying  to avoid too much manual settings/redundant information in LiteX-Boards to simplify things.
The I/O voltage will always the hardwired one the to the VCCIO of the bank. These additional constraints are in fact redundant with the IO Standards already defined in the Platform file and are just used by the tool for extra checks.

You can find more info at: https://stackoverflow.com/questions/65317854/how-to-change-pin-voltage-in-quartus-ii

For consistency with other board, I prefer not merging it for now but we'll see in the future it this has real additional value and should then do it for all platforms.

hansfbaier:
Oooooh, now I understand something. That explains a lot I have been struggling with. THANKS!

PR #258, opened 2021-08-17
lattice_crosslink_nx_evn: allow specifying the FPGA device

danc86:
This board is documented as having the LIFCL-40-9BG400C part, but some
versions of the board exist which were fitted with LIFCL-40-8BG400CES,
an engineering sample part. The distinction is important because the
engineering sample requires a different device ID to be embedded in the
bitstream. If you try to build a bitstream for LIFCL-40-9BG400C and load
it onto LIFCL-40-8BG400CES the configuration fails (indicated by the red
"INITN" LED on this board).

Accept --device to allow the user to specify which FPGA part their board
has.


enjoy-digital:
Thanks @danc86, that's merged.

PR #259, opened 2021-09-01
lattice_crosslink_nx_evn: don't set MASTER_SPI_PORT=SERIAL

danc86:
Setting MASTER_SPI_PORT=SERIAL causes the SPI flash pins to be reserved
for use by the sysCONFIG logic, and prevents user logic from assigning
them. This made it impossible to have a Litex design which accesses the
SPI flash on this board.

Remove the setting, so that we get the default behaviour which permits
user logic to assign these pins. In the unlikely event that someone
needs the pins to stay reserved for sysCONFIG after configuration (I'm
not sure why this would be needed) they could explicitly add this
command in their design.


enjoy-digital:
Thanks @danc86, in most cases we'll indeed probably want to favor the SPI Flash over sysConfig. Let's merge and we'll  rediscuss this if someone complains :)

danc86:
Thanks. I have a hunch that the only people in the world who have these boards, and are using Litex on them, are people I chat with and were also affected by this :-) but if indeed there was an important use for this then we can figure out something better.

enjoy-digital:
Indeed probably :)

PR #264, opened 2021-09-09
Add Alchitry Au as new board

teknoman117:
Add the Au board from Alchitry to LiteX.

System frequency is 83.333333 MHz, as the -1 speed grade xc7a35t can only manage 667 MHz DDR3L RAM.

Gated by https://github.com/enjoy-digital/litedram/pull/274 (timings for the RAM module on the board)

```
./alchitry_au.py --build --with-spi-flash --load
```


enjoy-digital:
Thanks @teknoman117, looks perfect!

PR #266, opened 2021-09-10
Add Mojo V3 support

teknoman117:
Add a board description for the (Alchitry / Embedded Micro) Mojo V3. Was able to get HDMI video out with the HDMI shield and some careful management of cache and SRAM sizes (the XC6SLX9 is sorely lacking in SLICEMs, which the FIFOs in litevideo really seem to want).

PXL_20210910_090931465


enjoy-digital:
Nice! Thanks @teknoman117 (and also thanks for making it very similar to other targets and adding notes).

PR #270, opened 2021-09-17
Add initial RZ-EasyFPGA support!

alainlou:
There is potentially one board issue which is noted in the #253 issue (specifically [this comment](https://github.com/litex-hub/litex-boards/issues/253#issuecomment-920558638)) and also [this code comment](https://github.com/litex-hub/litex-boards/compare/master...alainlou:master#diff-792c8ead031c2124b2580e48db8fbcc74cb5dbb89ba2f1496b5809527568513dR31)

Other than that, we have it working with:
- 4 LEDs
- UART
- External SDRAM

Thank you to these people for a ton of help and support with this so far!
@tcal-x @jevinskie

Usage:
`./rz_easyfpga.py --cpu-type=vexriscv --cpu-variant=minimal`

PR #272, opened 2021-09-23
rz_easyfpga: adjust SDRAM clk phase, also add 1:2 rate

alainlou:
Thanks @enjoy-digital for the pointer! You were right, the SDRAM clk phase did need to be adjusted. Adding wires had somehow mitigated this.

Reference: #270


enjoy-digital:
@alainlou: Great! This is merged.

PR #273, opened 2021-09-24
Wukong board improvements

ozbenh:
This adds support for v2 of the board via a --board-version argument
and a way to select the FPGA speed grade via a --speed-grade argument.

Note that the speed grade now defaults to -1. QMTech confirmed that
V1 of the board were made in two batches, one with -1 and one with -2,
while V2 of the board is all -1. So -1 is the safer default.

This also fixes the inversion of j10 and j11 and a typo in the pin
definition of jp3

Signed-off-by: Benjamin Herrenschmidt <benh@kernel.crashing.org>


enjoy-digital:
Thanks @ozbenh, that's merged.

PR #280, opened 2021-10-16
Fix tang nano loopback and button

troibe:
Before:
```
lxterm /dev/ttyUSB2

```

After:
```
lxterm /dev/ttyUSB2
jjj;lkj;lkj;lkj;lkj;lkj;lkj;lkj;lkj;lkjjjkjjjjjkl;lkjjjjjjkkkl;lkjjjj
```

Just got myself one of these as well. They could be a great for making learning about FPGAs more accessible.
Now that loopback is working...What would be the next steps?

### DIFF
--- a/litex_boards/platforms/sipeed_tang_nano.py
+++ b/litex_boards/platforms/sipeed_tang_nano.py
@@ -27,7 +27,7 @@ _io = [
 
     # Buttons.
     ("user_btn", 0, Pins("15"),  IOStandard("LVCMOS33")),
-    ("user_btn", 0, Pins("14"),  IOStandard("LVCMOS33")),
+    ("user_btn", 1, Pins("14"),  IOStandard("LVCMOS33")),
 
     # Serial
     ("serial", 0,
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Ah great @developandplay! I also wanted to look at it but hadn't found the time yet.

The next steps will probably to:
- Test UARTBone: `--uart-name=uartbone` and check if we can access the SoC's CSR (when generated with `--csr-csv=csr.csv`) with: `litex_cli --regs`, then `litex_cli --write` ... `litex_cli --read`.
- Integrate the external HyperRam.
- Integrate the LiteSPI core and test flashing of binary files with openFPGALoader.
- Try to build a SoC with a small CPU, external HyperRam and ROM code in SPI Flash with BIOS XiP (similar to Fomu/Icebreaker targets).

developandplay:
Hi Florent,
Thanks for the suggestions.
I tried the hyperram + serv yesterday but could only make it fit without
UART. So I couldn't really test if it's working as it should. Those
intermediate steps will be a great help with that.
Also do you have any other suggestions on what I could cut on the LiteX
side to save about 100 LUTs without sacrificing UART? I'll also ask Olof
about the serv side.
Best wishes
Martin

enjoy-digital ***@***.***> schrieb am Sa., 16. Okt. 2021,
09:09:

enjoy-digital:
I could have a closer look at it next week, but maybe you can try to reduce the fifo depth on the UART: https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/uart.py#L217-L218

developandplay:
Yes, that sounds good.
Olof explained (claimed?) that he was able to even fit multiple serv cores on this FPGA.
I wonder what would be the part of LiteX that eat up most space?
For UART Olof was using a bitbang approach.

PR #285, opened 2021-11-03
Use built-int RS232 port

lapnd:
Signed-off-by: lapd <lapd@soc.one>
The DE2-115 kit has built-in RS322 port. It's better to use that port than using extension slot


enjoy-digital:
Thanks @lapnd, that's merged.

PR #291, opened 2021-11-07
sqrl_acorn: add vivado programmer option, fix warnings

sergachev:
OpenOCD still struggles to program Artix flash when it's not the only device in a JTAG chain - but Vivado handles this easily, so I find the option useful.

CFGBVS and CONFIG_VOLTAGE should be correctly set in all designs according to Vivado.

### DIFF
--- a/litex_boards/platforms/sqrl_acorn.py
+++ b/litex_boards/platforms/sqrl_acorn.py
@@ -124,7 +124,9 @@ class Platform(XilinxPlatform):
         self.toolchain.bitstream_commands = [
             "set_property BITSTREAM.CONFIG.SPI_BUSWIDTH 4 [current_design]",
             "set_property BITSTREAM.CONFIG.CONFIGRATE 16 [current_design]",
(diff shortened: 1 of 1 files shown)


mithro:
@sergachev - Would you be interested in testing [OpenFPGALoader](https://github.com/trabucayre/openFPGALoader) and see how it compares to Vivado and OpenOCD?

sergachev:
@mithro: tried and it fails too. Board in question is https://github.com/enjoy-digital/litex-acorn-baseboard with ECP5 and Artix7 in chain. Diamond and Vivado work for respective chips flawlessly, so it's not a hardware issue. Should file issues to both OpenOCD and  OpenFPGALoader.

mithro:
@sergachev - Filing issues would be a great idea. I'm sure both OpenOCD and OpenFPGALoader want to be **more** reliable then Vivado / Diamond.

enjoy-digital:
Thanks @sergachev, this will indeed be useful to our use case on the baseboard.

PR #292, opened 2021-11-08
terasic_sockit now also can use the DRAM module from litedram

hansfbaier:
Now that the timings of the winbond chip has been verified working as expected,
we can use it here too.


enjoy-digital:

PR #293, opened 2021-11-15
Wrong pin / use default setup without CPU

hansfbaier:
This pin was wrong.
Now that the c library builds I could test the target.
Unfortunately I could not find a configuration which would fit the device.
Even SERV with 0x1000 sram would not fit (by a bit).
Since there is no hierarchial design I have no idea where to save some gates/mem.
So I disabled the CPU and just added an uartbone to be able to interact with the remaining components.

### DIFF
--- a/litex_boards/platforms/qmtech_10cl006.py
+++ b/litex_boards/platforms/qmtech_10cl006.py
@@ -29,7 +29,7 @@ _io = [
     ),
 
     # SDR SDRAM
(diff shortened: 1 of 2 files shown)


hansfbaier:
I tested SDRAM and leds via uartbone:
```bash
$ wishbone-tool -u /dev/ttyUSB0 0x40000000 0xc001babe
$ wishbone-tool -u /dev/ttyUSB0 0x40000000           
Value at 40000000: c001babe
```
of course interacting with leds works too.

enjoy-digital:

Issue #298, opened 2021-12-03
Butterstick: Use 2Gb DDR3 as default instead of 8Gb.


enjoy-digital:
"We have to inform you however that despite advertising a full gigabyte of DDR3 memory, the hardware in fact only has 2 gigabits of DDR3."

tpwrules:
As a temporary workaround, replacing MT41K256M16 with MT41K64M16 in https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/gsd_butterstick.py seems to work fine.

enjoy-digital:
With https://github.com/litex-hub/litex-boards/pull/299, 2Gb is now the default on Butterstick and 8Gb can be selected with `--sdram-device=MT41K512M16`.

PR #299, opened 2021-12-04
Butterstick updates

gregdavill:
I've added extra GPIO from the SYZYGY connectors.

Also related to #298 I've added `--sdram-device` to the cmd arguments, and set the default sdram-device to `MT41K64M16`


enjoy-digital:
Thanks @gregdavill!

PR #300, opened 2021-12-05
Better support for DE1-SoC

tilk:
The platform description for DE1-SoC was missing useful pin descriptions, including user LEDs, keys, switches, VGA output and GPIO ports. Also bitstream programming via `--load` did not work because of missing settings for `quartus_pgm`. This pull request fixes the issues described above.


enjoy-digital:
Thanks @tilk! The DE1-SoC support was indeed probably minimal, thanks for improving/fixing it.

PR #303, opened 2021-12-08
sipeed_tang_nano_4k: add option to build with Gowin EMCU

sergachev:
Add options to build with Gowin ARM hard CPU (--cpu-type=gowin_emcu). 
Software is not ready for ARM, so by default stays with vexriscv as before.

I changed options a bit so that vexriscv now requires --cpu-variant=minimal on command line - so that vexriscv is not hardcoded anymore, any CPU can be chosen from the command line.


enjoy-digital:
Great, thanks @sergachev, this is merged. Thanks for the `cpu-variant ` cleanup, I'm planning to do a pass on all targets to improve and uniformize this.

PR #311, opened 2021-12-19
Option --with-spi-flash for iCESugar-Pro

tilk:
The change is to use the same convention as in other boards: SPI Flash disabled by default, enabled by option.


enjoy-digital:
Thanks @tilk, this is merged.

PR #312, opened 2021-12-20
adding digilent_arty_z7 support

trabucayre:
this PR adds digilent arty Z7 (010 and 020 variant).
Tested as standalone FPGA and with processing_system and linux


enjoy-digital:
Great, thanks a lot @trabucayre. I also probably have it will do a test.

PR #313, opened 2021-12-21
efinix: add clock pins to t120 bga576 platform

fjullien:


enjoy-digital:
Thanks @fjullien, this is merged.

PR #314, opened 2021-12-22
LiteX BIOS on Zedboard

sergachev:
Uses "ZedBoard" Zynq IP preset built into Vivado for PS7 configuration (can be easily extended with Tcl if necessary).
Relies on https://github.com/enjoy-digital/litex/pull/1141 .
Required Xilinx C libraries (https://github.com/Xilinx/embeddedsw) are cloned into build directory during build.
Both SRAM and ROM regions are currently in actual DRAM address space - this is just a simplification for tests.
BIOS can be executed on Zynq using XIlinx xsct and JTAG with such a script: https://github.com/sergachev/litex-template/blob/dev/scripts/ps7_boot.tcl


enjoy-digital:
Thanks @sergachev! This is merged.

PR #315, opened 2021-12-27
Fix cannot find board 'littlebee'

hubmartin:
Seems like openFPGALoader capitalized letter 'B' in the past or made the boards case sensitive. Now flash works properly.

image


enjoy-digital:
Thanks @hubmartin, this is merged.

PR #319, opened 2022-01-06
antmicro-datacenter updates

kgugala:
This PR fixes clock pin location and extends eth reset time (required for the ksz9031 phy used on the board)


kgugala:
@enjoy-digital CI fails seem to be unrelated to the changes from this PR. Bunch of boards failed, but it is not clear why - the log in CI doesn't show much

enjoy-digital:
@kgugala: I'll look at this.

enjoy-digital:
Thanks @kgugala, this is merged. CI failing was related to some additional checks that we added to LiteX on Regions that I fixed with https://github.com/litex-hub/litex-boards/commit/8151bf7ffa8e224915d0e27d68e7e51236b71df3 (This was also a motivation to do some simplification on SPI Flash integration, so this fail has been useful :)).

PR #320, opened 2022-01-08
Added QMTech EP4CE55 board - almost identical to EP4CE15 board but bigger FPGA

robinsonb5:
The QMTech EP4CE55 board is very similar to the Cyclone IV board already in the repo - but has a much larger and more interesting FPGA.  (It's also used as the core of the Neptuno and UnAmiga Reloaded projects.)

### DIFF
--- a/litex_boards/platforms/qmtech_ep4ce55.py
+++ b/litex_boards/platforms/qmtech_ep4ce55.py
@@ -1,0 +1,157 @@
+#
+# This file is part of LiteX-Boards.
+#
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @robinsonb5, this is merged. If the only different is the FPGA device, we'll probably want to merge the platform/target in a common one (called ep4cex5?). I'll have a look at this soon.

enjoy-digital:
@robinsonb5: EP5CE55 and EP4CE15 platforms/targets have been merged with https://github.com/litex-hub/litex-boards/commit/45cfe2be6bb944278c6207b617fc60e33d6fade8.

PR #322, opened 2022-01-15
quicklogic_quickfeather: add eos_s3 arm software support

sergachev:
This enables LiteX BIOS run on EOS S3 ARM CPU.
The software library is adapted from QORC SDK examples for Quickfeather.
Because it's partially specific to Quickfeather (pin configuration etc) I propose to keep the library here in the boards repo for now.
Note that BIOS main() is started as a FreeRTOS thread - QORC SDK relies on FreeRTOS, maybe it could be decoupled later.

### DIFF
--- a/litex_boards/targets/libeos/Bootconfig.h
+++ b/litex_boards/targets/libeos/Bootconfig.h
@@ -1,0 +1,4 @@
+#ifndef __BOOT_CONFIG_H
+#define __BOOT_CONFIG_H
+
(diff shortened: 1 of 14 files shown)


enjoy-digital:
Thanks @sergachev, this looks good but I just revisited the libeos integration in #326 to attach it github for now until figure out the best place for it. This will avoid polluting litex-boards with this if this end up somewhere else.

PR #323, opened 2022-01-16
Digilent cmod a7: Fix litex build.

bl0x:
Meaning, that it fixed the build with symbiflow.


enjoy-digital:
Thanks @bl0x, this is merged.

PR #325, opened 2022-01-17
targets/digilent_nexys_video: add SATA generation argument

gsomlo:
allow SATA gen argument to be specified via command line


enjoy-digital:
Thanks @gsomlo, SATA Gen 3 will not be supported on the Nexys Video due to the -1 speedgrade, so I removed it with https://github.com/litex-hub/litex-boards/commit/06fb73ace36d3db3ef7479ca9b5d272d3c0bb46f.

PR #327, opened 2022-01-18
digilent_cmod_a7: Remove unused clocks.

bl0x:
This allows to also synthesize with 100 MHz instead of 48. Not sure, why these clocks are in. Possible copy and paste leftovers?
The cmod a7 does not have any DRAM, and no immediate need for serdes.

If these clock domains are some sort of litex convention, then please close.


enjoy-digital:
Thanks @bl0x, I did other cosmetic cleanups on top of it with https://github.com/litex-hub/litex-boards/commit/f11106e9c58c9ecd8f672a3169a2734107465b50.

PR #328, opened 2022-01-19
Fix Zedboard software

sergachev:
This cleans up Zynq memory map and removes a hack that is not necessary anymore.


enjoy-digital:
Thakns @sergachev, this is merged.

PR #333, opened 2022-01-21
xilinx_zcu106: add DDR4 interface and fix reset error

vacajk:
1. Add ZCU106 PL DDR4 PHY IOs. Then we can upload firmware to board for CFU playground framework.
2. According to the schematic diagram, it can be confirmed that the rst pin is active at high level.

DDR, reset pin and CFU tested pass.


enjoy-digital:
Perfect, thanks! This is merged.

PR #335, opened 2022-01-22
sipeed_tang_nano_9k: enable copackaged PSRAM

Icenowy:
Also enable SPI SDCard which was pending by the lack of main_ram
previously.

Signed-off-by: Icenowy Zheng <icenowy@aosc.io>

-----------------------------------
Depends on https://github.com/enjoy-digital/litex/pull/1176

### DIFF
--- a/litex_boards/platforms/sipeed_tang_nano_9k.py
+++ b/litex_boards/platforms/sipeed_tang_nano_9k.py
@@ -51,7 +51,15 @@ _io = [
         IOStandard("LVCMOS33"),
     ),
 
-    # TODO: SPI/RGB LCD, copackaged PSRAM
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @Icenowy, this is merged. Was the RAM working on hardware? (it can be tested with `mem_list` to get the list of memory regions and then `mem_test` on the region).

Icenowy:
@enjoy-digital well I initially initialized both chips, but then reverted on one only based on HW test result (because sometimes I get mem_test KO when adding two regions side-by-side)

maggo1404:
@Icenowy have you a idea for active the secend chip?

PR #336, opened 2022-01-23
Digilent CMOD A7: add flash support.

tcal-x:
Add both "--flash" and "--with-spi-flash"; tested on board.
4MB flash mapped at 0x00400000.

Signed-off-by: Tim Callahan <tcal@google.com>


enjoy-digital:
Thanks @tcal-x, that's merged.

PR #337, opened 2022-01-23
Enable LiteX BIOS on ARM core on Tang nano 4K

sergachev:
Has to be programmed with Gowin programmer (which for me sadly on works on Windows) in MCU mode using project.fs and bios.bin.

Also tested at "--sys-clk-freq=78e6" and "--sys-clk-freq=54e6".

### DIFF
--- a/litex_boards/platforms/sipeed_tang_nano_4k.py
+++ b/litex_boards/platforms/sipeed_tang_nano_4k.py
@@ -83,8 +83,8 @@ _connectors = [
 # Platform -----------------------------------------------------------------------------------------
 
 class Platform(GowinPlatform):
-    default_clk_name   = "clk27"
-    default_clk_period = 1e9/27e6
+    default_clk_name = "clk27"
+    default_clk_freq = 27e6
 
     def __init__(self):
         GowinPlatform.__init__(self, "GW1NSR-LV4CQN48PC7/I6", _io, _connectors, toolchain="gowin", devicename="GW1NSR-4C")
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @sergachev! This is merged. I however reverted some style changes with https://github.com/litex-hub/litex-boards/commit/1abb03e514840ddfbc7a541f88cba693d7a6fb09 with explanations. Each developer has different coding style, I'm just trying to keep things consistent between the different targets to simplify maintenance.

sergachev:
These style differences are the reason I opened https://github.com/litex-hub/ideas/issues/5 and still did not get any feedback.

My arguments are:
 - imports (at least unconditional) should be at the top - Style Guide for Python Code: https://www.python.org/dev/peps/pep-0008/#imports
 - 'M' is mega, 'm' is milli - so 'mB' is 1/1000 of a byte: https://en.wikipedia.org/wiki/Metric_prefix
 - duplication of definitions leaves more space for mistakes

enjoy-digital:
I saw litex-hub/ideas#5 and will answer when I'll have more time. For now style that you can find in other files should be used. Happy to discuss incoherencies but I also have the right to have preferences so:
- For imports, agree with you but the import that was was conditional and except on a few targets, SPI Flash is almost always optional.
- For M, if we change it, we should apply the change to all targets, I'll look at it.
- For changes in `default_clk_period`, as explain, I find it's putting the abstraction too far and don't necessarily want to use this.

sergachev:
Imports were unconditional before my changes - https://github.com/litex-hub/litex-boards/blob/5190c9c869ad61c9940a8b8a766b6808382e934c/litex_boards/targets/sipeed_tang_nano_4k.py - so sipeed_tang_nano_4k always required litespi and still does nearly always - except on gowin_emcu.

Sure, your right to set the style, just easier when it's clearly defined.

enjoy-digital:
Sorry if things are not clear, I'll try to improve this but have difficulty being reactive on all topics currently. I don't mind adapting the style while reviewing PR and am also listening to ideas for improvements. When possible improvements are identified that could apply to the project in general, I just have a preference to discuss them because it's easier to apply them to all targets at once.

PR #338, opened 2022-01-24
Add tweaks to Arty board to support yosys+nextpnr toolchain

suarezvictor:
This pull request requires the related one: https://github.com/enjoy-digital/litex/pull/1180

Example command for testing (NOTE: some parameters are *required* to make it work as the clock frequency and memory setting):
`./digilent_arty.py --sys-clk-freq=50e6 --integrated-main-ram-size=8192 --cpu-type=femtorv --toolchain=yosys+nextpnr --build`

PATH needs to point to `/prjxray/utils/` to find required scripts

the "Misc()" attributes seems not to affect only the previously defined signals so it's not accepted in yosys+nextpnr toolchain

### DIFF
--- a/litex_boards/platforms/digilent_arty.py
+++ b/litex_boards/platforms/digilent_arty.py
@@ -295,7 +295,7 @@ def sdcard_pmod_io(pmod):
             Subsignal("mosi", Pins(f"{pmod}:1"), Misc("PULLUP True")),
             Subsignal("cs_n", Pins(f"{pmod}:0"), Misc("PULLUP True")),
             Subsignal("miso", Pins(f"{pmod}:2"), Misc("PULLUP True")),
-            Misc("SLEW=FAST"),
+            Misc("SLEW=FAST"), #NOTE: this is not supported by yosys+nextprn toolchain
             IOStandard("LVCMOS33"),
         ),
         ("sdcard", 0,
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @suarezvictor, this is merged.

enjoy-digital:
I just did some minor changes on top with https://github.com/litex-hub/litex-boards/commit/621d45cd9e15a5c4d95fcfe1dd7572738185468a to avoid some of the yosys+nextpnr specific code. I'll try to test it on hardware soon and will maybe try move the specific code to the build generation (ex: avoid generating the Misc, INTERNAL_VREF constraints to the `.xdc` with a warning).

suarezvictor:
Good to see you were able to do more tweaks that are required and better handled in your hands :)

In regards to the "Misc" they may be supported but with different way of generation. That needs more review but maybe there is a better solution than just deletion.
When try to you test it in hardware, you'll see the instantiation of FD primitives should be avoided too.

links for tool installation:  https://github.com/BrunoLevy/learn-fpga/blob/master/FemtoRV/TUTORIALS/toolchain_arty.md
requires this tweaks:  BrunoLevy/learn-fpga#54
all this inspired from: https://github.com/gatecat/nextpnr-xilinx/tree/xilinx-upstream/xilinx/examples/arty-a35

PR #340, opened 2022-01-26
digilent arty z7: allows toolchain selection (PL only)

trabucayre:
Tested with arty Z7 10 (xc7z010) & arty Z7 20 (xc7z020).

zynq7000 variant not supported due to xilinx's specifics commands required to generates the CPU macro template.


enjoy-digital:

PR #342, opened 2022-01-27
Get the default IP prefix from LITEX_DEFAULT_IP_PREFIX env-var if set.

jevinskie:


enjoy-digital:
Thanks @jevinskie, sorry this is complicating things for me and makes code less readable. We could re-discuss this together. For now I prefer not merging.

PR #344, opened 2022-02-06
Add 85F to help for orangecrab device

benstobbs:
The 85F orangecrab board exists, and works fine when this option is set to 85F, but leaving it out of the help is a bit confusing.


enjoy-digital:
Thanks @benstobbs for catching this, this is merged.

PR #345, opened 2022-02-06
Add Xilinx KV260 support

sergachev:
The configuration is sufficient to run LiteX BIOS on Cortex-A53 core #0.

Can be executed on hardware using JTAG with a following [xsct script](https://github.com/sergachev/litex-template/blob/kv260/scripts/psu_boot.tcl); `make -f Makefile.kv260 load` in https://github.com/sergachev/litex-template/tree/kv260 will build everything and run `xsct` in the end.

Relies on https://github.com/lucaceresoli/zynqmp-pmufw-builder to create a generic PMU firmware; first build will take a while because it includes a cross-toolchain.

### DIFF
--- a/litex_boards/platforms/xilinx_kv260.py
+++ b/litex_boards/platforms/xilinx_kv260.py
@@ -1,0 +1,34 @@
+#
+# This file is part of LiteX-Boards.
+#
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @sergachev! This is merged.

PR #348, opened 2022-02-08
Draft: Add Lattice ECP5 VIP board + HDMI output board support

hubmartin:
WIP: will do some cleanup soon
Let me know if there is something more like code style or something else I need to fix.
Thanks

https://twitter.com/hubmartin/status/1490430338266451975

### DIFF
--- a/litex_boards/platforms/lattice_ecp5_vip.py
+++ b/litex_boards/platforms/lattice_ecp5_vip.py
@@ -1,0 +1,144 @@
+#
+# This file is part of LiteX-Boards.
+#
+# Copyright (c) 2019 Arnaud Durand <arnaud.durand@unifr.ch>
+# SPDX-License-Identifier: BSD-2-Clause
+
+from litex.build.generic_platform import *
+from litex.build.lattice import LatticePlatform
+from litex.build.lattice.programmer import OpenOCDJTAGProgrammer
+
+import os
+
+# IOs ----------------------------------------------------------------------------------------------
+
+_io = [
+    # Clk / Rst
+    ("clk27",  0, Pins("E17"), IOStandard("LVCMOS33")), ##
+    ("clk100", 0, Pins("C5"), IOStandard("LVDS")), ##
+    ("ext_clk50",    0, Pins("B11"), IOStandard("LVCMOS33")),
+    ("ext_clk50_en", 0, Pins("C11"), IOStandard("LVCMOS33")),
+    ("rst_n",        0, Pins("AH1"),  IOStandard("LVCMOS33")),##
+
+    # Leds
+    ("user_led", 0, Pins("AG30"), IOStandard("LVCMOS25")),
+    ("user_led", 1, Pins("AK29"), IOStandard("LVCMOS25")),
+    ("user_led", 2, Pins("AK30"), IOStandard("LVCMOS25")),
+    ("user_led", 3, Pins("AH32"), IOStandard("LVCMOS25")),
+    ("user_led", 4, Pins("AG32"), IOStandard("LVCMOS25")),
+    ("user_led", 5, Pins("AJ29"), IOStandard("LVCMOS25")),
+    ("user_led", 6, Pins("AM28"), IOStandard("LVCMOS25")),
+    ("user_led", 7, Pins("AM29"), IOStandard("LVCMOS25")),
+
+    # ws2812
+    ("ws2812", 0, Pins("AK31"), IOStandard("LVCMOS25")),
+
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @hubmartin, happy to fix style/CI. Can you however update the copyrights and add yours?

hubmartin:
Thanks @enjoy-digital for your help.

I've cleaned a code a bit, removed some comments and WS2812 test code.
It is ok to have Twitter handle as a copyright contact? I could also point to my website where the email could be found. I just don't feel like giving my email in the wild.

Before finalizing this merge, I might try to improve some issues I had, maybe you could help me:

1) Not sure if the DDR3 is correct, mainly frequency, maybe it could be faster? (maybe related to video issue below). I just copied stuff from other ECP5/DDR3 baord, changed pins and it somehow magically started to work. Kit datasheet w schematics and part numbers https://www.latticesemi.com/-/media/LatticeSemi/Documents/UserManuals/E---I-Folder-3/FPGA-EB-02001-1-7-ECP5-VIP-Processor-Board.ashx?document_id=52101
2) I edited `sdram_test` in BIOS code to test the whole 256 MB of DDR3 RAM and it passed. Does it mean that memory/pinout is 100% validated?
3) Video terminal is working at 800x600@60, however when run video terminal at 1920x1080@60 I see strange artifacts in the image. Like some color shift or font-rom issue with shifted purple color like 8 or 16 pixels to the right. Could this be because DDR is clocked too slow? More in this thread https://twitter.com/hubmartin/status/1490337242652590084
image

Thanks for any ideas.

hubmartin:
4) (Hopefully) Last DDR3 thing - The memory interface is 32bit on this kit, not sure if this mens something to `"1:2"` DDR rate, I did not fully understand what that rate means. I copyied the DDR code from `lattice_versa_ecp5` platform which has 16bit memory bus.

enjoy-digital:
Thanks @hubmartin, the current support seems good enough for merging, we could discuss/improve things in the future.
Using your twitter handle is also perfectly fine if preferred. To answer your different questions:
1) If working, it should be very close to be correct. I'll have a closer look when I'll have more time and will do a closer review. Since there are no CDC between the CPU and DRAM currently, frequencies of each others are tightly linked/correlated and the CPU frequency will probably here limit the maximum DRAM frequency.
2) If `mem_test` pass on the full DRAM range, pinout should indeed be correct.
3) This would need to be investigated, it's also possible that at this resolution timings are not met. It would be interesting to increase progressively the resolution and see up to which resolution this artifact is not present. From what I remember seeing on Twitter, this artifact is also present with the official bitstream?
4) `1:2` is the DRAM rate ratio, this means that the controller runs with a 1x Clk while PHY runs with a 2x Clk. This is used to find a good compromise between CPU freq/DRAM freq without requiring CDC as explained in 1. Here LiteX will automatically be able to adjust the internal data-width from a different external data-width, so it's good and with 32-bit you'll have 2X the bandwidth available than on the Versa with 16-bit :)

hubmartin:
Thank you for merging and answering all the questions.

>From what I remember seeing on Twitter, this artifact is also present with the official bitstream?
It was my initial suspection, but it seemed like the official bitstream used pink pixels to show some edge/area detection. 

I was able to run "1024x768@60Hz" at 65e6 Hz, but I'm not able to generate "1280x720@60Hz" frequency 74.25e6 with error `ValueError: No PLL config found`. I thought that ECP5 has many PLLs which could generate precise frequencies. Might this be the issue of my pll code and should I somehow initiate second PLL? I'm a bit lost in all the clock tree dependecies and clock domains.

So for my "1920x1080@60Hz" (with that purple artifacts) which requires 148.5e6 Hz pixel clock I used 160e6 Hz to get it running, but there might be also an issue in the timing. So I would start with precise clock first.

Thanks for any help.

hubmartin:
I gave it a shot.
I tried instantiate a new PLL and the correct frequency is created, the display menu shows exactly 60 Hz and 67.3 kHz
Not sure if this is right approach.
image

However the artifacts stayed the same
IMG_20220219_183255

hubmartin:
With that second PLL I can confirm that 1280x720@60Hz works fine (74.25e6 Hz)

I tried "1920x1080@30Hz" with 89.01e6 Hz but none of my three HDMI displays/TVs does not support this framerate and suggests 60 Hz.

With "1920x1080@60Hz" I tried to increase CPU freq from default 60 MHz in increments of 20 MHz up to 120 MHz and system was stable and running fine, but the display artifacts we're still visible.

PR #353, opened 2022-02-21
Add Xilinx ZCU216 support

sergachev:
Very similar in all ways to KV260 - BIOS runs on Cortex-A53 core #0, Wishbone <-> AXI works to control LEDs in the logic.


enjoy-digital:
Thanks @sergachev, this makes it probably the most expensive board currently supported in LiteX-Boards :)

PR #354, opened 2022-02-21
Add initial support for RCS Arctic Tern boards

madscientist159:
Add initial (pre-release) support for the upcoming Arctic Tern ECP5-based modules and their associated BMC development carrier card.


madscientist159:
@enjoy-digital Thank you, and we will! :smile:

PR #357, opened 2022-02-28
Add ColorLight i9 v7.2

DaveBerkeley:
The ColorLight i9 v7.2 is very similar to the i5 v7.0.

It has an ECP5 -45F instead of the -25F. The user LED is on a different pin. It has a different SPI Flash part.

I've modified the colorlight_i5.py files to take a board="i9" parameter :+1: 

    --board i9 --revision 7.2

See https://github.com/wuxx/Colorlight-FPGA-Projects/blob/master/colorlight_i9_v7.2.md for the hardware description.


enjoy-digital:
Thanks @DaveBerkeley, this looks good! This is merged.

PR #359, opened 2022-03-01
antmicro datacenter: fix pins and increase freq

acomodi:
This PR fixes the antmicro datacenter board definition:

 - adjusted pins corresponding to the RDIMM datacenter board
 - increased clk frequency to 100 MHz to reach 800 MT/s
 - use A7DDRPHY (without odelays)
 - add i2c


enjoy-digital:
Thanks @acomodi, this looks good and is merged.

PR #361, opened 2022-03-03
adi_adrv2crr: Add support for the ADI ADRV2CRR with ADRV9009-ZU11EG SoM

smunaut:
This is a carrier board with a SoM mounted on it.
There is also an FMC connector that can accept another
AD-FMCOMMS8-EBZ to get two more ADRV9009 RFIC but support for
that is not added yet.

Note that the PCIe support requires :
 - Change the .xci in the litepcie to use the right Quad
 - Revert litex 3c34039b731b42e27e2ee6c8e399e5eb8f3a058f so the
   timing constrainst of litepcie apply correctly

Signed-off-by: Sylvain Munaut <tnt@246tNt.com>

### DIFF
--- a/litex_boards/platforms/adi_adrv2crr_fmc.py
+++ b/litex_boards/platforms/adi_adrv2crr_fmc.py
@@ -1,0 +1,499 @@
+#
+# This file is part of LiteX-Boards.
+#
+# Copyright (c) 2022 Sylvain Munaut <tnt@246tNt.com>
+# SPDX-License-Identifier: BSD-2-Clause
+
+from litex.build.generic_platform import *
+from litex.build.xilinx import XilinxPlatform
+from litex.build.openocd import OpenOCD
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @smunaut, as discussed, we'll try to generate the .xci automatically in the future, similarly to 7-Series, so this should simplify things for this hardware and avoid the manual change.

https://github.com/enjoy-digital/litex/commit/3c34039b731b42e27e2ee6c8e399e5eb8f3a058f has been introduce to allow the build on some boards. I'll investigate to see which ones and will try to see how we can cover all cases (or if we add a switch to fhis).

smunaut:
Another possible option for that commit would be to split the `LOC` constraints and the timing constrains into two different xdc files.  The `LOC` one could be processed early and the timing one using the normal.

PR #363, opened 2022-03-08
add Gowin programmer support

curliph:
1. update only target for sipeed_tang_nano_9k platform and target
2. update LiteX to support PowerShell and WSL host for tang_nano_9k
3. add gowin programmer support, but with limitations:
    - there is no details info for external spi programming with programmer_cli, so it's support not working by now.
    - Verify got stuck, so it's also disabled by default.
    - programming interface is compatible with openFPGALoader

4. test with oss-cad-sute (from yosysHQ) under Powershell and also WSL(ubuntu 20.04)


enjoy-digital:
Thanks @curliph, this is merged.

PR #366, opened 2022-03-11
targets/nexys-video: Add support for sata pll refclk

gsomlo:


enjoy-digital:
Thanks @gsomlo, this is merged.

PR #371, opened 2022-03-16
ZCU102 Target file

jwfaye:
Hello everyone,

I submit the supports of ZCU102 board, I have not test the basecore yet  as I want some advises(I am downloading the vivado toolchain though). There is some stuff I do not really understand in the support of the ZCU104 board (I will open an issue about that). To create the new target file, I compare the simple target et the ZCU104 target implementation to create a basic one for the 102 board, 

What do you think about it ? 

Jo


enjoy-digital:
Thanks @jwfaye, the target has already been merged in previous PR.

PR #372, opened 2022-03-22
butterstick: set ethernet rx_delay to 0ns

mkj:
The Microchip KSZ9031RNX PHY on the Butterstick has a default 1.2ns
internal RX delay so we shouldn't add the default 2ns MAC delay.

In testing with Linux on vexriscv I haven't seen any difference either
way, but with liteeth in Microwatt I have seen 30%+ packet loss when
receiving from certain ethernet devices (RTL8153 and AX88179 usb-gige
adapters, a GS105 switch didn't show the problem). Setting RX delay=0
resolves the problem. A TX delay is still required by the PHY.

Adding @gregdavill for any thoughts.

### DIFF
--- a/litex_boards/targets/gsd_butterstick.py
+++ b/litex_boards/targets/gsd_butterstick.py
@@ -129,7 +129,9 @@ class BaseSoC(SoCCore):
         if with_ethernet or with_etherbone:
             self.submodules.ethphy = LiteEthPHYRGMII(
                 clock_pads = self.platform.request("eth_clocks"),
-                pads       = self.platform.request("eth"))
+                pads       = self.platform.request("eth"),
+                rx_delay   = 0e-9, # KSZ9031RNX phy adds a 1.2ns RX delay
+                )
             if with_ethernet:
                 self.add_ethernet(phy=self.ethphy, dynamic_ip=eth_dynamic_ip)
             if with_etherbone:


gregdavill:
Interesting that some devices result in higher packet loss. My understanding was that the RX/TX PHY delays happen at a later stage in the pipeline, a fair distance from the MDI, but maybe skew between pairs can close down the receiving signal eye?

This seems like an appropriate change, my main testing on the ethernet with Butterstick r1.0 has just been if a link was established, and packets back and forth, so likely wouldn't have noticed this level of packet loss.

On an old Butterstick r0.1 project I've used the PHYs delay adjustments to map out and test how close to the ideal capture point in the eye we are. Seems like it might be worth re-running that on the Butterstick.

It would just be nice to check if 0ns is the ideal value. Or if it should be closer to the theoretical ~0.8ns?

mkj:
What were you using to check ideal values, litescope? I can have a go at that - be interesting to see the difference between devices.

There was a step change in loss between 127 and 128 byte length frames.

gregdavill:
I had an SoC running that enabled loopback mode in the PHY, then it tried sending a packet and reading it. I'd log results and create a histogram of packet errors vs delay value.

enjoy-digital:
@mkj @gregdavill: I'm merging since from your results it's already improving behavior with some equipment.  We could adjust this in the future if a closer review show that it's not the ideal theoretical value.

PR #375, opened 2022-03-23
ADRV2CRR: Fixing some of my screwups that I just noticed now

smunaut:
I had never tried to use some of the devices that were on the board. Attempting to do so today showed some mistakes in the platform file.


enjoy-digital:
Thanks @smunaut, that's merged.

PR #376, opened 2022-03-24
qmtech_5cefa2: make serial pins consistent with other boards

hansfbaier:
The pins on this one are inconsistent with the other qmtech boards.


enjoy-digital (approved):
Thanks @hansfbaier!

PR #379, opened 2022-03-26
[enh] added efinix t8f81 dev kit

chmousset:
Requires https://github.com/enjoy-digital/litex/pull/1254 to load or flash

Built with
```
python ./litex_boards/targets/efinix_t8f81_dev_kit.py --no-uart --build --load
```


enjoy-digital (approved):
Thanks @chmousset! This looks good.

PR #383, opened 2022-04-12
Added Kintex-7 xc7k420t xc7k420tiffg901-2L named as u420t board

sysmanalex:
Added Kintex-7 xc7k420t xc7k420tiffg901-2L named as u420t fpga board.
verified

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -85,3 +85,4 @@ Copyright (c) 2020      Vamsi K Vytla <vamsi.vytla@gmail.com>
 Copyright (c) 2021      Vegard Storheil Eriksen <zyp@jvnv.net>
 Copyright (c) 2020      YanekJ <yanekx@gmail.com>
 Copyright (c) 2021      Yoshimasa Niwa <niw@niw.at>
+Copyright (c) 2021      Alex Petrov <sysman@sysman.net>
--- a/README.md
+++ b/README.md
@@ -199,6 +199,7 @@ Some of the suported boards, see yours? Give LiteX-Boards a try!
     ├── trenz_te0725
     ├── trenz_tec0117
     ├── tul_pynq_z2
+    ├── u420t
     ├── xilinx_ac701
     ├── xilinx_alveo_u250
     ├── xilinx_alveo_u280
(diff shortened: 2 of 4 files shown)


enjoy-digital:
Thanks @sysmanalex! I'm going to merge it and will probably do minor cosmetic changes. I also think I'll add an aliexpress prefix in the name of these boards because they all seem to come from the same (but unknown...) manufacturer. Is the u420t the official name or one you gived it to?

enjoy-digital:
I did some changes on top of your PR to simplify things: https://github.com/litex-hub/litex-boards/commit/b2a346edc849500c367edce8ce31b755a7ef636d.

sysmanalex:
Sure. Name change is up to you.
there is no tags on boards, and no clue about designer - btw provided
schematics contains some mistakes.
1) I was thinking about Wiki for all boards with photos, I can host and
support web-site if you like or we may create separate github wiki.
please answer here yes or no if you support it.
2) my boards from aliexpress - my order was over 1.5 year ago, shop is
closed - but similar shops still selling here
4001151287611.html
1005001775106297.html
price was 150-170$ now ~250$

aliexpress sellers also sell very similar old design board
"Xilinx Kintex-7 XC7K420T FPGA Supercomputer Machine Learning Algorithms
Verify Power Board"
32959014756.html - 200-250$ depends from locations / EU 230-240EUR
32979417274.html
33027545095.html

and NEW design, looks nice and clear a-la QMTECH level, but price +70$
1005003142524650.html

enjoy-digital:
Thanks for the info on the board. I just renamed the platform/target to `aliexpress_xc7k420t.py`.

Regarding the board description/photos, I'm just generally trying to include a link to some available online documentation for boards that are not well known.  Maintaining the LiteX support for the boards is already taking lots of time, so I don't want to create documentation for the boards myself. If someone wants to do it, we could add a link in the README, but otherwise I think it's better to just reuse already existing documentation.

sysmanalex:
a) there will be a lot of different xc7k420t boards, on AliExpress ~ 30
models, pcie, qmtech, alinx, mk, kc, chuanglong, de7420.
b) AliExpress is too long name, then you need to wrote it all the time &
boards often sells via alibaba, ebay, tens of china/eu shops imho.
c) reason why I suggested wiki, we can list all boards, assign short names
and keep in single location all photos, schematics and pdf docs for boards.

Short and uniq names makes life much simpler then litex boards definitions
list will be hundreds+

Some schematics, specs, very hard to find or it can be as well in Chinese

Just suggestions.

PR #388, opened 2022-05-03
muselab: icesugarpro: fix HDMI output

swetland:
- enable data_n outputs
- use drive_both property of updated VideoHDMIPHY to drive both
  differential outputs
- adjust default clocking to successfully close timing
  (sysclk at 50Mhz, HDMI at 25MHz, HDMIx5 at 125MHz)


swetland:
This depends on: https://github.com/enjoy-digital/litex/pull/1290

swetland:
It was suggested (thanks, tnt) that a simpler fix would be to adjust the IOStandard of the pins, but it turns out that in the case of the IceSugar Pro, the pins were not chosen well:
ERROR: cannot place differential IO at location PIOB

PR #391, opened 2022-05-15
Move HDMI Support From Sipeed Tang Nano 4K.

minexo79:


enjoy-digital:
Thanks @minexo79! Has it been validated on hardware?

minexo79:
Yep, but actually I'm use CFU Playground to test it, not use Vanilla LiteX Environment.

enjoy-digital:
@minexo79: Ok thanks, so we can merge.

PR #393, opened 2022-05-19
muselab icesugar pro: add ethernet/etherbone support

swetland:
- use P4 header for a waveshare ethernet phy module
- add --with-ethernet, --with-etherbone, --eth-ip, and --eth-dynamic-ip
  target configuration options


enjoy-digital:
Thanks @swetland, that's merged.

PR #398, opened 2022-06-13
a_e115fb: new board

Icenowy:
It's a core board with EP4CE115 by a random vendor on Taobao.

Signed-off-by: Icenowy Zheng <uwu@icenowy.me>


enjoy-digital (approved):
Thanks @Icenowy! This looks good. I'm going to merge and will also probably just add a `taobao` prefix as I did for boards from `aliexpress` (not perfect but since vendor is unknow, this still clarify things a bit).

PR #404, opened 2022-06-27
target: basys3: Remove redundant sdcard additions

robertszczepanski:
This removes a redundancy that prevents a target to build with `--with-sdcard` and `--with-spi-sdcard` parameters.


enjoy-digital:
Thanks @robertszczepanski!

PR #406, opened 2022-06-28
[fix] instanciate PLL for valentyUSB

chmousset:
Currently, building icebreaker_bitsy with --uart-name usb_acm does not work.
Here we instantiate the PLL to create the two clocks required by valentyusb and force sys_clk to 48MHz, if usb_acm is selected.

tested on v1.1 hardware with:
```python
python litex_boards/targets/icebreaker_bitsy.py --cpu-type serv --build --flash --csr-csv build/icebreaker_bitsy/csr.csv --uart-name usb_acm
python litex_boards/targets/icebreaker_bitsy.py --cpu-type serv --build --flash --csr-csv build/icebreaker_bitsy/csr.csv --uart-name serial --sys-clk-freq 60e6
```

### DIFF
--- a/litex_boards/targets/icebreaker_bitsy.py
+++ b/litex_boards/targets/icebreaker_bitsy.py
@@ -34,7 +34,9 @@ mB = 1024*kB
 # CRG ----------------------------------------------------------------------------------------------
 
 class _CRG(Module):
-    def __init__(self, platform, sys_clk_freq):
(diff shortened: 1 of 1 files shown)


tcal-x:
Hi @chmousset , were you able to communicate with the BIOS over usb/ACM using `litex_term`?

PR #407, opened 2022-06-29
targets/sipeed_tang_primer_20k: don't force run when build argument is true

trabucayre:
With sipeed_tang_primer_20k `builder.build()` `run` argument is set to true when `--build` is true too.
Usually, for all others boards, the rule is to let Builder deciding when  `run` must be true according `--no-compile` or `--no-compile-gateware` are set / unset.


enjoy-digital:
Ah sorry, just fixed it before seeing the PR @trabucayre :) (with https://github.com/litex-hub/litex-boards/commit/190f272c14ebcc0b38faaaec7c2da7e4c76280aa). So not merging but thanks!

PR #409, opened 2022-07-11
Remove DM signals assignment from DDR4 Datacenter platform

RRozak:
Data Masks (DM) are unsupported on x4 devices and wrong pins were assigned.


enjoy-digital:
Thanks @RRozak, this is merged.

PR #410, opened 2022-07-14
added new boards ld_krote and ld_schoko

machdyne:
This adds support for the Lone Dynamics Kröte FPGA dev board and the Lone Dynamics Schoko FPGA computer.

https://github.com/machdyne/krote
https://github.com/machdyne/schoko


enjoy-digital:
Thanks @machdyne! This looks good. Regarding the vendor's name, it would be better to avoid an abbreviation. I imagine Lone Dynamics is too long, but why not just using machdyne since this "Machdyne™ is the hardware division of [Lone Dynamics Corporation](https://lonedynamics.com/)." and since this is where you are selling the boards?

machdyne:
I went ahead and made those changes, there is also a pending PR for linux-on-litex-vexriscv where I've changed the prefix as well. Please let me know if I need to make any further changes. Thanks @enjoy-digital !

enjoy-digital:
Thanks @machdyne for the update, this is merged.

PR #414, opened 2022-08-11
lambdaconcept_ecpix5: fix HDMI pin

povauboin:


enjoy-digital:
Good catch, thanks @povauboin!

PR #415, opened 2022-08-17
digilent_arty: make GPIOs interrupt-capable if SoC has IRQs enabled

lschuermann:
Enables interrupt requests for the exposed GPIOs if the SoC is capable of handling interrupts.


enjoy-digital:
Thanks @lschuermann!

PR #416, opened 2022-08-24
Add Antmicro Artix DC-SCM board

shenki:
Based on commits by:

  - Jedrzej Boczar <jboczar@antmicro.com>
  - Karol Gugala <kgugala@antmicro.com>

Tested serial and networking on a DC-SCM with mor1k and vexriscv.


shenki:
We're doing some work with this board so I would like to put it in the repository.


FYI @jedrzejboczar @kgugala

enjoy-digital:
Thanks @shenki, this looks good to me.

PR #418, opened 2022-08-26
sipeed_tang_primer_20k: adding dock lite support

trabucayre:
@sipeed provides two docks for tang primer 20k core board: one called simply **dock** (expected for october) and **lite** version with:
- 2 buttons
- 2 switches
- 4 PMODs
- 2 2 x 20 headers

This PR adds the definition for *io* and *connectors* and an argument to select between docks version. Default dock is left unchanged, user has to explicitly set `--dock lite`.
It also move serial node at core level: **lite** has no onboard serial adapter (*rx* and *tx* are available only on core board **J1** connector), it disable inconditionnaly **led_chaser** because there is no leds at all.

Also toolchain is explicitly set at target level and it fix *CARD1* typos.

### DIFF
--- a/litex_boards/platforms/sipeed_tang_primer_20k.py
+++ b/litex_boards/platforms/sipeed_tang_primer_20k.py
@@ -19,6 +19,13 @@ _io = [
     # Clk / Rst.
     ("clk27",  0, Pins("H11"), IOStandard("LVCMOS33")),
 
+    # Serial.
+    ("serial", 0,
+        Subsignal("rx", Pins("T13")), # CARD1:1
+        Subsignal("tx", Pins("M11")), # CARD1:11
+        IOStandard("LVCMOS33")
+    ),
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @trabucayre! This looks good. For the connectors on the Lite Dock, we could also express then in function of CARD1. Not sure this will be currently supported, but this is something that could be interesting to add in the future to cascade connectors.

trabucayre:
Indeed: with this board and, also,  **qmtech** core boards and futur the **ULX4M** this make sense to have a recursive pin resolution.
It's currently not supported: ASAP I will look at this (and adding something similar to `add_extension` to avoid having to [extend](https://github.com/litex-hub/litex-boards/blob/master/litex_boards/platforms/sipeed_tang_primer_20k.py#L255) connectors list)

PR #419, opened 2022-08-31
targets/sipeed_tang_primer_20k: fix CLKDIV (disable reset and calibration)

trabucayre:
`CLKDIV` primitive must have `RESETN` and `CALIB` set to an explicit value. Without that nothing is produce on output.
By fixing that issue HDMI is now working.
Tested with **lite** dock, a @machdyne [DDMI PMOD](https://machdyne.com/product/ddmi-pmod/) and with colorbars and terminal.


enjoy-digital:
Good catch, thanks @trabucayre! (Will test tomorrow on the standard dock).

PR #422, opened 2022-09-10
Adding DFU Support to the Butterstick Platform

azzeloof:
I ran into an issue while trying to get LiteX running on my Butterstick, it was pointed out to me that the example expects a JTAG adapter for programming the board, rather than the onboard USB. I added the option to select between "jtag" and "dfu" while programming so that the board can be programmed without an additional adapter. If no programmer option is specified, it defaults to a JTAG programmer, to maintain compatibility with existing code.


enjoy-digital:
Thanks @azzeloof! This merged.

PR #427, opened 2022-09-25
aliexpress_stlv7325: Fix missing parameter for PULLUP attribute

pftbest:


enjoy-digital:
Thanks @pftbest, good catch.

PR #429, opened 2022-09-28
targets/litex_acorn_baseboard: fix pn_swap

trabucayre:
With this [PR](https://github.com/enjoy-digital/litex/pull/1442) focused on fixing this [issue](https://github.com/enjoy-digital/litex/issues/1439) blue and red colors are swapped. This fix as for consequence the need to update all targets where `pn_swap` is used.
Currently two targets uses this option `sipeed_tang_primer_20k` and `litex_acorn_baseboard`. Former has no needs to be updated since all channels must be swapper but latter has only channels 1 and 2 to permut -> this PR fix that by replacing **b** by **r** (aka channel 2).


enjoy-digital:
Thanks @trabucayre!

PR #430, opened 2022-10-01
stlv7325: Adjust DDR3 pins to match the reference design

pftbest:
Fixes memory corruption when using 64bit CPUs like Rocket or NaxRiscv.

See https://github.com/enjoy-digital/litedram/issues/285#issuecomment-1263886574


enjoy-digital:
Thanks @pftbest, good to know this fix memtester issues.

PR #431, opened 2022-10-03
MNT RKX7: update platform and target for D-2 release

mntmn:
This is an update for the MNT RKX7 files to make it compatible with the release version (tested in MNT Reform motherboard/carrier). Details: https://source.mnt.re/reform/reform-kintex-som

Hardware Changes from D-1:

- The DDR3 memory routing was completely redone for a less congested layout and avoiding crosstalk. This fixed some address bus errors of the first version.
- The TC358778XBG MIPI encoder was removed. Instead, 24-bit parallel RGB, clocks and extra GPIOs are routed out to a new 50 pin FPC header. This makes the module more flexibly usable.
- We designed a new adapter PCB featuring the Analogix ANX9807 RGB to eDP chip. We are able to successfully drive MNT Reform's eDP Full HD display.
- The TUSB1310ZAY USB3/2 PHY was removed as it is EOL, expensive and there was is no available XHCI host implementation that could drive it. Instead, USB1/2 pads are now driven directly by the FPGA. This works well with the SpinalHDL OHCI host controller which is supported by LiteX.
- 8 FPGA additional pads were broken out to an array of square test points.

### DIFF
--- a/litex_boards/platforms/mnt_rkx7.py
+++ b/litex_boards/platforms/mnt_rkx7.py
@@ -2,6 +2,7 @@
 # This file is part of LiteX-Boards.
 #
 # Copyright (c) 2021 Florent Kermarrec <florent@enjoy-digital.fr>
+# Copyright (c) 2022 Lukas F. Hartmann, MNT Research GmbH <lukas@mntre.com>
 # SPDX-License-Identifier: BSD-2-Clause
 
 from litex.build.generic_platform import *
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Great work @mntmn! Thanks for the update/PR.

enjoy-digital:
@mntmn: for info, LiteX is providing classical video timings but more specific/exotic ones can be provided externally. I just changed integrated your specific video timings in the target file with https://github.com/litex-hub/litex-boards/commit/d0dd009329e968100e87c630a1cc54b3017b91af.

PR #437, opened 2022-10-20
targets/redpitaya: fix csr & reset region

trabucayre:
redpitaya board fails to build when --cpu-type is set to **zynq7000**. This is
due to a missing reset address and `csr` region.
This PR fix this issue by adding the **flash** and **csr** region.
It also disable the uart and set `integrated_ram_size` to 0.


enjoy-digital:
Thanks @trabucayre, this is merged.

PR #438, opened 2022-10-21
platforms/nexys4*: Update part name

shawnanastasio:
Symbiflow/f4pga don't recognize the part name xc7a100t-CSG324-1, so change it to xc7a100tcsg324-1 which works with both f4pga and Vivado.


mithro:
@shawnanastasio - Please log a bug on f4pga too, it should recognize the same part names as Vivado.

enjoy-digital:
Thanks @shawnanastasio, as said by @mithro, this should probably be fixed at F4PGA level.

enjoy-digital:
Merging since we are using this format for some Xilinx platforms. This would still be good to have the different format supported in F4PGA and do some cleanup in LiteX-Boards to make things consistent between platforms.

PR #443, opened 2022-10-28
targets/digilent_arty_z7: adding software support

trabucayre:
This PR adds the `finalize` method to fetch **embeddedsw** repository, fills `xparameters.h` (used to provides UART to use and RAM size). These steps are required to build the **litex bios** for this board.

Tested with **xcst**:
```
xsct -nodisp ps7_boot.tcl \
        $PS7_INIT_TCL $MAIN_ELF $BIT_FILE
```
where:
- `PS7_INIT_TCL`: the ps7_init.tcl: *build/TARGET/gateware/TARGET.srcs/sources_1/ip/TARGET/ps7_init.tcl*
- `MAIN_ELF`: the bios file
- `BIT_FILE`: the bitstream

ps7_boot.tcl contains:

```
connect
targets -set -filter {name =~ "ARM*#0"}
rst
stop
source [lindex $argv 0]
ps7_init
ps7_post_config

dow [lindex $argv 1]
fpga [lindex $argv 2]
con

```

Side note:
- maybe adding an arg to disable the **bios** build when user want to use bootloader+linux instead of baremetal
- code added by this PR is really similar to zedboard: refactorize may make sense to avoid code dupplicate.

### DIFF
--- a/litex_boards/targets/digilent_arty_z7.py
+++ b/litex_boards/targets/digilent_arty_z7.py
@@ -12,6 +12,7 @@ from litex_boards.platforms import digilent_arty_z7
 from litex.build import tools
 from litex.build.xilinx import common as xil_common
 from litex.build.xilinx.vivado import vivado_build_args, vivado_build_argdict
+from litex.build.tools import write_to_file
 
 from litex.soc.interconnect import axi
 from litex.soc.interconnect import wishbone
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks a lot @trabucayre; this will provide a good example (and I should also try it :)). This is merged; we should also get the information from the PR in a wiki page in the future.

trabucayre:
wiki -> true. I've spent lot a time yesterday to find where I have stored my docs :-/

PR #445, opened 2022-10-30
stlv7325, a_e115fb: use the proper vendor name Sitlinv

Icenowy:
The boards are in fact from a vendor called 成都赛特凌威科技有限公司, and their English registered trademark (used on the banner of their Taobao store) is Sitlinv, which sounds like 赛特凌威.

Use this vendor name instead of where it's bought.

Signed-off-by: Icenowy Zheng <uwu@icenowy.me>


enjoy-digital (approved):
Thanks a lot @Icenowy!

PR #450, opened 2022-11-13
Some enhancements to STLV7325 target code

Icenowy:


enjoy-digital:
Thanks @Icenowy! That's merged.

Icenowy:
> Thanks @Icenowy! That's merged.

Should we make some of these code more common?

e.g. the local-ip/remote-ip parameters could work for all boards with Ethernet, Jtagbone parameter could workd for all compatible FPGAs, etc.

enjoy-digital:
Yes, that's something we've been discussing recently with @trabucayre while trying to simplify the targets but haven't had time to had a look yet. We should have some code in LiteX able to fill the arguments and interpret them to configure the cores, but I haven't really look how to do this. If you want to explore this, happy to discuss.

PR #455, opened 2022-11-23
fix: Basys3 V_sync is pin R19

guissalustiano:
I was running VGA on my Basys 3 board and it wasn't working
Debugging I realized that the Vsync pin was wrong and changing it worked

We can see the correct pin [on contrains](https://github.com/Digilent/digilent-xdc/blob/master/Basys-3-Master.xdc#LL129C34-L129C37) or in [page 11 of docs](https://digilent.com/reference/_media/basys3:basys3_rm.pdf)
image


enjoy-digital:
Thanks @guissalustiano!

PR #457, opened 2022-11-29
RFC: Don't Apply: vc707: wire up "missing" DDRAM pins

gsomlo:
There was originally a "ddram_dual_rank" record available among the list of io options for, among others, xilinx vc707, added by commit [373e74f4 ("add new board files for VC707 and VCU118...")](https://github.com/litex-hub/litex-boards/commit/373e74f435da8ed16592e3f9d416eca4fa70b656)

The "ddram_dual_rank" entry was duplicating some of the pins used by the "ddram" record, but I assume is intended to suppport using SODIMMs larger than the default 1GB included with the board?

Subsequently, commit [c093d0d0 ("platforms: cleanup pass to uniformize comments/separators/orders.")](https://github.com/litex-hub/litex-boards/commit/c093d0d0fc13165c74c67a024a66f94764a60594) removes this option.

@enjoy-digital, @fei-g -- any idea what the real story behind "ddram_dual_rank" is (is my assumption about it being for support of larger SODIMMs correct)? Any idea what the ***largest*** SODIMM is that can be supported? Any way to make this a configurable option (add back "ddram_dual_rank" as a aseparate option, or somehow enable larger, non-"standard" SODIMM sizes in the `target` vc707 file)?

Please treat this PR as an actual *issue* for now, don't apply until we understand what the best available option actually is.

Thanks!


mithro:
The team form @antmicro doing DDR memory related stuff might know? @kgugala any thoughts?

enjoy-digital:
Hi @gsomlo,

the Xilinx dev boards generally allow Dual Rank SODIMMs but are equipped  with Single Rank SODIMM by default -> That's the reason I removed the Dual Rank definition when doing the cleanup. 

If you use the Dual Rank IOs definition with LiteX and a Single Rank SODIMM, it will not work, since LiteDRAM will try to use the second Rank. One possibility if we want to keep a single definition of the IOS would be to create an equivalent  [PHYPadsReducer](https://github.com/enjoy-digital/litedram/blob/master/litedram/common.py#L82-L105) but for ranks.

Dual Ranks are generally larger but one other interesting aspect is that it also allow doubling the number of banks on the controller (since each rank is independent), increasing the hit rate and then performance a bit.

For more info, you can look at https://www.hardwaretimes.com/single-rank-vs-dual-rank-memory-which-one-is-better-for-gaming/ for example.

@antmicro are working on making LiteDRAM parameters dynamic and allowing re-configuration from settings read from the SPD EEPROM (the main motivation is to be able to easily switch SODIMM modules on [RowHammer-Tester](https://github.com/antmicro/rowhammer-tester) project). This feature will also probably be very useful for the Xilinx boards with SODIMMs since will also allow easy switch of SODIMM and a next step could also be to handle Single/Dual Ranks also dynamically.

gsomlo:
@enjoy-digital thanks much for the information!

When I tried building LiteDRAM for vc707 with the second Rank pins patched in, the port width went from 256 bits to 512, and the amount of expected memory also doubled to 2GB (as expected).

The thing that still confuses me is that I read reports about the board being able to support ***4GB*** of total memory, using e.g. a [MT8KTF51264HZ-1G9](https://media-www.micron.com/-/media/client/global/documents/products/data-sheet/modules/sodimm/ktf8c128_256_512x64hz.pdf) chip. Looking at the (linked) data sheet, the 4GB chip does not appear to require more pins, so I'm wondering how 4GB could be made to work with LiteDRAM (assuming that's possible and I haven't missed some minute difference between the pinout used by the 2GB vs. the 4GB SODIMM variant). Any clue much appreciated!

gsomlo:
I think I now (mostly) understand how things hang together. I'll open a new issue or PR once I have a chance to run some tests on actual hardware. Thanks again for helping me think through it!

PR #458, opened 2022-12-03
Arty z7 tcl config

trabucayre:
Currently **digilent_arty_z7** target fetch a tcl script from somewhere. This file contains almost all parameters usefull and useless. Also, each time this targets is build `wget` fetch again the *tcl* file resulting on a long list of **arty_zXXX.tcl.X** with *X* tending to infinite...
This PR replace this call by a dict with only parameters required for the configuration. Since z10 and z20 variants differs only by the PL size the same config is working in all cases.

This PR also adds a note on how to load gateware+bios (unfortunately it's using proprietary tool).


enjoy-digital:
Thanks @trabucayre!

PR #463, opened 2022-12-28
terasic_deca: add SPI SD card support

stone3311:
This pull request adds support for the integrated SD card socket on the DECA.

The board uses the SN74AVCA406L level shifter as an interface between the FPGA and the SD card. This means that some additional signals are necessary to set the direction of the command and data pins.


enjoy-digital:
Thanks @stone3311!

PR #467, opened 2023-01-06
Allow building digilent_arty using f4pga

### LINKED ISSUE #466: digilent_arty fails to build using f4pga due to dependency on unsupported features (DNA_PORT, XADC)
Building digilent_arty.py with `--toolchain vivado` succeeds, but when built with `--toolchain f4pga` the build fails with the following error:

```
Found constant-zero generator '$abc$82080$auto$blifparse.cc:518:parse_blif$89937.T0'
# Load circuit took 0.35 seconds (max_rss 90.7 MiB, delta_rss +32.2 MiB)
The entire flow of VPR took 1.21 seconds (max_rss 90.7 MiB)



stderr:
Error 1: 
Type: Blif file
File: /home/osboxes/src/synth/build/digilent_arty/gateware/digilent_arty.eblif
Line: 184330
Message: Failed to find matching architecture model for 'DNA_PORT'
```

Per the https://github.com/chipsalliance/f4pga/issues/648 thread, this issue occurs because the DNA_PORT block is unsupported by f4pga.

Commenting out https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/digilent_arty.py#L97 results in a similar error about the 'XADC' block, which is also unsupported by f4pga.

Once both the call to `DNA()` and `XADC()` are commented out, then a bitstream can be successfully generated with the f4pga toolchain and is able to be booted on the device itself.

mlaga97:
Only use XADC() and DNA() functions if vivado is the current toolchain.

I believe that this fixes #466 and https://github.com/chipsalliance/f4pga/issues/648

### DIFF
--- a/litex_boards/targets/digilent_arty.py
+++ b/litex_boards/targets/digilent_arty.py
@@ -91,11 +91,13 @@ class BaseSoC(SoCCore):
         SoCCore.__init__(self, platform, sys_clk_freq, ident="LiteX SoC on Arty A7", **kwargs)
 
         # XADC -------------------------------------------------------------------------------------
-        self.xadc = XADC()
+        if toolchain == "vivado":
+            self.xadc = XADC()
 
         # DNA --------------------------------------------------------------------------------------
-        self.dna = DNA()
-        self.dna.add_timing_constraints(platform, sys_clk_freq, self.crg.cd_sys.clk)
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, we can do this yes and remove it when f4pga will support these features.

PR #480, opened 2023-03-09
Update sipeed_tang_nano_9k.py

maggo1404:
add pin info for _connectors


enjoy-digital:
Thanks @maggo1404! That's merged.

PR #482, opened 2023-03-13
sipeed_tang_nano_20k: new board

Icenowy:
This board uses Gowin GW2AR series chip (which is GW2A with integrated RAM).

Support for the integrated SDRAM on Tang Nano 20K is still TODO.

Note: currently when the SD card is enabled, block 0 could be correctly read but block 1 will fail.

==================

Depends on https://github.com/enjoy-digital/litex/pull/1646 .


enjoy-digital:
Nice, thanks @Icenowy! Let's merge this and improve/fix the issues progressively. I should also receive a Tang Nano 20K soon so could also do some testing.

PR #488, opened 2023-04-03
sitlinv_stlv7325: use S7PLL instead of S7MMCM for system clock

Icenowy:
As we do not need fine phase tweaking for the main system clock, use S7PLL instead of S7MMCM to allow higher VCO frequency and more flexible sys_clk_freq.


enjoy-digital:
Thanks @Icenowy, this is merged.

PR #489, opened 2023-04-03
targets/nexys-video: simplify sata configuration

gsomlo:
Make nexys-video SATA configuration similar to the way it is supported on the majority of other SATA-capable targets.


enjoy-digital:
Thanks @gsomlo, this is merged. The refclk selection was useful in the past to do the bringup together on your hardware IIRC, but we can indeed avoid it in LiteX-Boards. This is still present in the LiteSATA bench (https://github.com/enjoy-digital/litesata/blob/master/bench/nexys_video.py) since can be useful for test purposes.

PR #493, opened 2023-04-11
make DDR3 memory work on the DECA with the UniPHY

hansfbaier:
This is now a working setup for DDR3 with UniPHY on the DECA board.


enjoy-digital:
Great thanks!

PR #495, opened 2023-04-17
Fix Memory initialization of Alveo U200 failed #1606

vietthanh85:
I fixed the issue Memory initialization of Alveo U200 as discussed in https://github.com/enjoy-digital/litex/issues/1606.
Both offnaria and I confirmed that it works OK.


enjoy-digital:
Thanks @vietthanh85, that's merged.

PR #498, opened 2023-04-22
Initial support for STLV7325 (v2) Kintex-7 board.

gsomlo:
This is the 2nd (2023) version of the board sold through https://www.aliexpress.us/item/3256801088848039.html

Link to the vendor-provided schematic: [XC7K325TFFG676-2原理图.pdf](https://github.com/litex-hub/litex-boards/files/11301409/XC7K325TFFG676-2.pdf)

### DIFF
--- a/litex_boards/platforms/sitlinv_stlv7325_v2.py
+++ b/litex_boards/platforms/sitlinv_stlv7325_v2.py
@@ -1,0 +1,510 @@
+#
+# This file is part of LiteX-Boards.
+#
+# Copyright (c) 2023 Gabriel Somlo <gsomlo@gmail.com>
+# Copyright (c) 2022 Andrew Gillham <gillham@roadsign.com>
+# SPDX-License-Identifier: BSD-2-Clause
+
+from litex.build.generic_platform import *
+from litex.build.xilinx import Xilinx7SeriesPlatform
+from litex.build.openocd import OpenOCD
+
+# This board is available here:
+# https://www.aliexpress.com/item/1005001275162791.html
+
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @gsomlo! Do you know if there are lots of differences between the two revisions of the boards or if only a few IOs that are changes? If the latter, this could be worth trying to merge v1 and v2 platforms/targets in a single file as we are doing for some other boards (ex OrangeCrab, ColorLight, etc...).

gsomlo:
@enjoy-digital -- I force-pushed an updated version. Have a look (diff the platform files and see how many pins are different, and how you feel about trying to make things conditional vs. just using a separate pair of board & target files).

Off the top of my head:
- the 100MHz clock is now a 50MHz clock
- gmii ethernet is now rgmii
- hdmi has fewer connected pins
- external BTB interface has a bunch of additional pins connected
- many of the interfaces have *some* of the pins changed from v1

LMK what you think.

enjoy-digital:
Hi @gsomlo,

this new revisions indeed seems quite different. I'm fine with a different platform/target then since merging the two will make things more difficult to follow/understand.

gsomlo:
RGMII Ethernet, UART, microSD card, and (1G included) DDRAM successfully tested and working with litex.

Not working: SATA

Not tested: SFP, HDMI, FMC/LPC, other external connectors

NOTE: on FMC/LPC, ***none*** of the `DP0*` pins are connected to the FPGA, which precludes using an ***external*** SATA adapter such as e.g. the [AB09-FMCRAID](https://www.dgway.com/AB09-FMCRAID_E.html)

PR #506, opened 2023-05-27
Add spiflash definition to linsn_rv901t board

sensille:
Add signal definition for spiflash signals as described here:

https://github.com/q3k/chubby75/blob/master/rv901t/doc/hardware.md#connector-jp5-spi-flash


enjoy-digital:
Thanks @sensille.

PR #513, opened 2023-07-31
lattice_ecp5_evn: add_jtagbone flag

josuah:
This follows https://github.com/enjoy-digital/litex/pull/1087 which allows using the built-in JTAG for both the FPGA programming and the internal designs. Kuddos!

Starting LiteX server with it:
```
$ litex_server --jtag-config=targets/openocd.cfg --jtag
Open On-Chip Debugger 0.12.0+dev-g7023deb (2023-07-29-00:21)
Licensed under GNU GPL v2
For bug reports, read
        http://openocd.org/doc/doxygen/bugs.html
jtagstream_serve
Info : ftdi: if you experience problems at higher adapter clocks, try the command "ftdi tdo_sam
ple_edge falling"
Info : clock speed 25000 kHz
Info : JTAG tap: ecp5.tap tap/device found: 0x81113043 (mfg: 0x021 (Lattice Semi.), part: 0x111
3, ver: 0x8)
Warn : gdb services need one or more targets defined
[CommUART] port: JTAG / tcp port: 1234
Connected with 127.0.0.1:40662
Disconnect
Connected with 127.0.0.1:40674
Disconnect
Connected with 127.0.0.1:46370
Disconnect
Connected with 127.0.0.1:44228
Disconnect
```

The content of `openocd.cfg` (separate pull request with that):
```
adapter driver ftdi
adapter speed 25000
transport select jtag
ftdi vid_pid 0x0403 0x6010
ftdi channel 0
ftdi layout_init 0x00e8 0x60eb
reset_config none

set _CHIPNAME ecp5
jtag newtap $_CHIPNAME tap -irlen 8 -expected-id 0x81113043
```

And then, I can get the bus values!
```
litex@lap1:$ litex_cli --regs --csr-csv=csr.csv
0xf0000000 : 0x00000000 ctrl_reset
0xf0000004 : 0x12345678 ctrl_scratch
0xf0000008 : 0x00000000 ctrl_bus_errors
0xf0001000 : 0x00000000 leds_out
0xf0001800 : 0x00e4e1c0 timer0_load
0xf0001804 : 0x00000000 timer0_reload
0xf0001808 : 0x00000001 timer0_en
0xf000180c : 0x00000001 timer0_update_value
0xf0001810 : 0x00000000 timer0_value
0xf0001814 : 0x00000001 timer0_ev_status
0xf0001818 : 0x00000001 timer0_ev_pending
0xf000181c : 0x00000000 timer0_ev_enable
[...]

PR #514, opened 2023-08-01
OpenOCD config file updates

josuah:
As discussed in #513 here are two commits adjusting the config:

- 63adcb931b2c44c0b4c3d8a28cde2732acc1c318 which fixes a bug when `$_CHIPNAME` is not defined:
  `set _CHIPNAME ecp5` as the only change.

- 95b6fa1587612bda7279bfdcb1e7b843db5f1c76 which convert the syntax to the supported format.

These kind of refactoring are not *expected* to break anything, but still has the risk of any refactoring.
Should I keep the second commit away until everything can be tested again?

### DIFF
--- a/litex_boards/prog/openocd_butterstick.cfg
+++ b/litex_boards/prog/openocd_butterstick.cfg
@@ -1,10 +1,10 @@
-interface ftdi
-ftdi_vid_pid 0x0403 0x6014
-ftdi_channel 0
-ftdi_layout_init 0x00e8 0x60eb
+adapter driver ftdi
+adapter speed 25000
+transport select jtag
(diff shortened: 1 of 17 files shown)


enjoy-digital:
Thanks @josuah, we are still using an old version of OpenOCD (https://github.com/enjoy-digital/openocd) and we'll have to switch to upstream before merging (at least test it on my side and evaluate the changes required for some other projects). I'll try to do it in not too long, but probably not before the end of the month.

josuah:
I am actually glad that this gets deferred as it means it costs not more than it helps.
I will be so very patient... :)

josuah:
Big refactors are best done by the authors themselves, that way no pressure either direction :)

PR #515, opened 2023-08-02
Allow use of OpenOCD for the Crosslink-NX

josuah:
This adds the support for Crosslink-NX EVN FTDI in JTAG mode, through OpenOCD.
This does not yet permit to use the JTAG for a serial connection with the bus (JTAGBone), as the `JTAG` instance is still not figured out, but hopefully I'll get there soon!


enjoy-digital:

PR #519, opened 2023-08-09
Crosslink-NX EVN: fix UARTBone always disabled

josuah:
On 5412d0e I introduced a bug, I think by committing a debug branch instead of the code I tested with.

While at it, convert `register_mem` (deprecated) to `bus.add_slave`.

I tested again as described on 5412d0e with the same results.


enjoy-digital:

PR #524, opened 2023-08-18
targets/lattice_crosslink_nx_evn: add main_ram section for firmware

josuah:
This takes the values from the Antmicro SDI MIPI converter as a model and is enough to run a Zephyr hello world, but not seemingly enough for a the Zephyr Shell sample.

Related: https://github.com/litex-hub/zephyr-on-litex-vexriscv/pull/13


enjoy-digital:

PR #526, opened 2023-09-01
Added QMTECH RP2040 Daughterboard

Chandler-Kluser:
Added new QMTECH Daughterboard with RP2040, which can be found in [AliExpress](https://www.aliexpress.com/item/1005005094654777.html).

It has some differences compared to the previous model.

Documentation can be found [here](https://github.com/ChinaQMTECH/DB_FPGA_with_RP2040).


litex_boards/targets/qmtech_rp2040_daughterboard.py
```diff
@@ -0,0 +1,136 @@
...
+            11: "J2:52",
+            12: "J2:51",
+            13: "J2:50",
+            14: "J2:49",
+            15: "J2:48",
+            16: "J2:47",
+            17: "J2:46",
+            18: "J2:45"
+        }),
+    ]
```
josuah, line 136: Often, the list of I/Os and connectors go to a separate "platform/...py" source file.

Is it intended to keep it here?

Chandler-Kluser: I did not know, I will drop another commit

josuah: Maybe moving `litex_boards/targets/qmtech_rp2040_daughterboard.py` to `litex_boards/platforms/qmtech_rp2040_daughterboard.py` is enough? As there is not really possible to run this script and get a board out of it.

Chandler-Kluser: yes, I am doing that!

Chandler-Kluser (commented):
Fixed Issue

enjoy-digital:
Thanks @Chandler-Kluser, that's merged.

PR #532, opened 2023-09-25
Resolve High Density bank IOStandard error

lianakoleva:
This resolves the following error during `build` on Vivado 2023.1:
``ERROR: [DRC BIVB-1] Bank IO standard Support: Bank 47 has incompatible IO(s) because: The LVDS I/O standard is not supported for banks of type High Density. Move the following ports or change their properties:
clk125_p``


enjoy-digital:
Thanks @lianakoleva!

PR #535, opened 2023-10-05
qmtech_daughterboard.py: fix package for importing WaitTimer

hansfbaier:


enjoy-digital:
@hansfbaier: This has been done to start decoupling LiteX from Migen, so is expected.

PR #537, opened 2023-10-13
xilinx_zcu102: Add pin definitions for DDR4 SDRAM and FMC connectors, add litedram to the target.

rniwase:


enjoy-digital:
Thanks @rniwase!

PR #540, opened 2023-10-19
platforms/sipeed_tang_primer_25k: swap UART TX & RX, fix TX pin (J1:20 -> J1:21)

trabucayre:


enjoy-digital:
Thanks @trabucayre.

PR #541, opened 2023-10-23
QMTech boards: make Artix7 bitstream/platform commands conditional on toolchain

hansfbaier:
I also removed setting .f4pga_device since it is identical to platform.device

### DIFF
--- a/litex_boards/platforms/qmtech_artix7_fbg484.py
+++ b/litex_boards/platforms/qmtech_artix7_fbg484.py
@@ -161,16 +161,17 @@ class Platform(Xilinx7SeriesPlatform):
             connectors += daughterboard.connectors
 
         Xilinx7SeriesPlatform.__init__(self, device, io, connectors, toolchain=toolchain)
+
         self.toolchain.bitstream_commands = \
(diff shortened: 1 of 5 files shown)


enjoy-digital:
Thanks @hansfbaier. In the past, I tried to avoid specializing platforms/target to the requirements of the different toolchain. We agree to allow it on the Digilent Arty to still allow fast experimentation/developments but with the aim in the long term to remove it. I think the different toolchain will try to align with Vivado commands in the long terms,  so it would probably be better to handle there unsupported commands directly in the LiteX build backend for the toolchain. Ex have the build backend catch a `"set_property INTERNAL_VREF"` command and generate a Warning since not supported but also not inserting it in the generated files. This way, as the toolchain support improve, you'll just be able to modify the build backend and progressively remove the modifications, without requiring modifying the different platforms.

hansfbaier:
Yes that totally makes sense. Can you point me to the code location where that would be?

hansfbaier:
OK, implemented the requested changes here.

enjoy-digital:
Thanks @hansfbaier.

PR #544, opened 2023-10-23
Improve terasic_de2_115 target support

madprogrammer:
This PR adds additional peripherals to the Terasic DE2-115 board support files, such as 2 MII Ethernet ports, SD card and on-board LEDs


enjoy-digital:
Thanks @madprogrammer! This is merged.

PR #559, opened 2024-01-13
sipeed_tang_mega_138k: Added PCIe reset and other pins from sipeed dock

racerxdl:
I added few more pins (that I use for some projects) to tang-mega platform.

I also added `use_ready_as_gpio` and `use_done_as_gpio` to toolchain options, which allows to use two LEDs in the SOM board as well (here mapped as `led_done` and `led_ready`)


enjoy-digital:
Thanks @racerxdl!

PR #560, opened 2024-01-27
AXAU15: Updated FMC+ pinout

Johnsel:
There were some pins wrong/missing in the FMC+ definition.


enjoy-digital:
Thanks @Johnsel!

PR #564, opened 2024-02-14
Update sipeed_tang_nano_9k.py so serial works

blark:
I wanted to save folks the three hours of troubleshooting why there was no serial terminal that I went through yesterday :) I confirmed that with these lines uncommented I cannot connect to the device, when I comment them out the terminal/bios works fine.

Commenting these lines out was the advice I found in a number of issues. I am not sure why it's not in the board file already, I figured I'd try and help out.

### DIFF
--- a/litex_boards/targets/sipeed_tang_nano_9k.py
+++ b/litex_boards/targets/sipeed_tang_nano_9k.py
@@ -73,7 +73,7 @@ class BaseSoC(SoCCore):
 
         # SoCCore ----------------------------------------------------------------------------------
         # Disable Integrated ROM
-        kwargs["integrated_rom_size"] = 0
+        # kwargs["integrated_rom_size"] = 0
         SoCCore.__init__(self, platform, sys_clk_freq, ident="LiteX SoC on Tang Nano 9K", **kwargs)
 
         # SPI Flash --------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


trabucayre:
Hi,
As mentioned in another issues, tang nano 9k FPGA is quite small: shipping CPU firmware is an extra cost (RAM, FF, ...). Of course it's possible to do like that but free space, for extra cores, wil be near to 0...
Have you tried to write the firmware using external flash (as required):
```
openFPGALoader --external-flash -o 0 build/sipeed_tang_nano_9k/software/bios/bios.bin
```
?

Thanks

enjoy-digital:
Hi @blark,

Thanks for the PR and sorry for the late feedback.

The issue you hit is real, but the change is not something we should make the default for the Tang Nano 9K target.

This target intentionally disables the integrated ROM and boots the BIOS from external SPI flash. The Tang Nano 9K is quite small, and keeping the BIOS in FPGA fabric consumes  resources that are usually needed for the actual design. If only the bitstream is loaded to SRAM and the BIOS has not been programmed to external flash, the CPU will reset to an   empty/invalid flash region, so the serial BIOS prompt will not appear.

The expected flow is to also program the BIOS to external flash, for example with openFPGALoader:

 ```sh
  openFPGALoader --external-flash -o 0 build/sipeed_tang_nano_9k/software/bios/bios.bin
  ```
or use the target’s --flash flow, which handles the bitstream and BIOS programming for this board when using openFPGALoader.

So I don’t think we should merge this as-is, since it changes the default boot/resource tradeoff.

enjoy-digital:
Closing in favor of #751. The issue reported here is valid, but this patch changes the Tang Nano 9K default from external-flash BIOS boot to integrated-ROM boot, which costs scarce FPGA resources. #751 keeps the current external-flash default and adds an explicit --with-integrated-rom option for SRAM-only loading/debug.

PR #566, opened 2024-02-27
Qmtech k325: fix wrong button assignment

hansfbaier:


enjoy-digital:
Thanks @hansfbaier!

litex_boards/platforms/qmtech_xc7k325t.py
```diff
@@ -16,6 +16,10 @@
     # Clk / Rst
     ("clk50", 0, Pins("F22"), IOStandard("LVCMOS33")),
 
+    # Buttons
+    ("user_btn_n", 0, Pins("AF9"),   IOStandard("LVCMOS33")),
```
minghan1, line 20: Should the IOStandard be LVCMOS18? LVCMOS33 causes an error during the place_design stage in Vivado:
"[DRC BIVB-1] Bank IO standard Support: Bank 33 has incompatible IO(s) because: The LVCMOS33 I/O standard is not supported for banks of type High Performance. Move the following ports or change their properties: user_btn_n0."

In the QMTech example, LVCMOS18 is used. [QMTech Example Link](https://github.com/ChinaQMTECH/QMTECH_XC7K325T_CORE_BOARD/blob/main/Software/Test01_led_key.zip)

hansfbaier: Yes you are right. Both user buttons are 1.8V. I checked the schematic. Please submit a PR.

PR #568, opened 2024-03-02
Enable bitstream compression on Kintex K325T

hansfbaier:
I forgot to enable this one!
Important for such a large part.


enjoy-digital:
Thanks.

PR #571, opened 2024-03-08
Correct pinout for VGA connector

skrutt:
I spent quite some time scratching my head why i got no picture from litex when all looked good and the default basys example produced a very crisp picture.. Until i found this!

And an open question: i notice there are no defines either for the onboard flash or the 4x7 segment, should i add these? perhaps in a separate PR?


enjoy-digital:
Thanks @skrutt, this is merged. Adding SPIFlash and 7 Segment support would indeed be a nice other addition in another PR yes.

PR #573, opened 2024-03-22
Fix sqrl_acorn proxy bitstream for cle-101

Xilokar:
The cle-101 (and the litfury) is a xc7a100t, but we were using the wrong proxy bitstream when flashing


enjoy-digital:
Thanks @Xilokar!

PR #578, opened 2024-03-30
qmtech core boards: make core resources optional

hansfbaier:
This makes derivative designs (like MiSTeX) more flexible.


enjoy-digital:
Thanks @hansfbaier.

PR #583, opened 2024-04-22
alientek davincipro: fix speedgrade

hansfbaier:
Yes, it is actually speed grade 2:
image


hansfbaier:
Now the PCIe error is gone, but it says it can't fit the design on the A35T:
```
ERROR: [DRC UTLZ-1] Resource utilization: RAMB18 and RAMB36/FIFO over-utilized in Top Level Design (This design requires more RAMB18 and RAMB36/FIFO cells than are available in the target device. This design requires 116 of such cell types but only 100 compatible sites are available in the target device. Please analyze your synthesis results and constraints to ensure the design is mapped to Xilinx primitives as expected. If so, please consider targeting a larger device.)
```
I tried halving the dram cache, but that didn't have an impact. Will try halving again.

enjoy-digital:
Good, thanks @hansfbaier.

PR #586, opened 2024-05-17
platforms/xilinx_zcu106.py: added connectors and FMC HPC0/1

trabucayre:
This PR has for goal to add `_connectors` part and complete with the two FMC available on this board


enjoy-digital:
Thanks @trabucayre!

PR #589, opened 2024-06-13
efinix_trion_t20: add sdram

maass-hamburg:
add sdram of the efinix trion t20 bga256 dev kit.

The sdram will only work, if the system clock frequency is not bigger than 50 MHz.
When the spiflash is also activated, the sdram will fail partly with any system frequency of 50 MHz, that's why the default had been lowered to 45 MHz, because at that frequency it works.


enjoy-digital:
Thanks @maass-hamburg!

PR #590, opened 2024-06-19
ZynqXXX boards: remove CSR definition and GP0 connection to CPU

trabucayre:
With this [PR](https://github.com/enjoy-digital/litex/pull/1994) CSR are now defined at CPU core level and the first CPU interface is automatically connected to the SoC. This allows to reduces logic at target level and behavior is now nearest as others CPUs.

Tested using zedboard, artyZ7 and ZCU102.


enjoy-digital:

PR #596, opened 2024-07-05
platforms/xilinx_vcu118: add fmc and pmod connector pinouts

gaoyichuan:
VCU118 board has one FMC HPC, one FMC+ HSPC and two PMOD connectors, Vivado board files does not contain pinouts for these, so pinouts are taken from VCU118 Evaluation Board User Guide (UG1224).


enjoy-digital:
Thanks @gaoyichuan!

PR #598, opened 2024-08-01
enclustra mercury kx2+: fix pinnumber on connector C

malled2002:
FPGA Pin U17 is on connector C:112 instead of C:113 according to the [kx2+ schematics](https://download.enclustra.com/public_files/FPGA_Modules/Mercury+_KX2/Mercury-KX2-R2-2_User_Schematics_V2.pdf)


enjoy-digital:
Thanks @malled2002.

PR #602, opened 2024-08-14
platforms/sipeed_tang_primer_20k.py: fix IOStandard values

trabucayre:
As mentioned in *led*'s comment  `IOStandard` have a wrong value.
This PR fix the value for all resources.
It also disable and adds a not for *i2c* pins for HDMI and LCD due to a conflict with ethernet.


enjoy-digital:
Thanks @trabucayre.

PR #603, opened 2024-08-16
make Gowin boards work with Apicula

pepijndevos:
Thanks to https://github.com/enjoy-digital/litex/pull/2036 by @Mai-Lapyst LiteX supports Apicula with the Himaechel target now, but it appears that the boards do not yet work correctly.

This is a WIP PR to try and fix those. I figured I'd put it out there to avoid duplicating work with Mai.

So far I've been mostly threading `toolchain`  through to the platform. Currently it fails to connect the RAM chips with their SERDES and SiP IO stuff. Some of that may be a more long-term project.

Depends on https://github.com/enjoy-digital/litex/pull/2042

### DIFF
--- a/litex_boards/targets/sipeed_tang_nano.py
+++ b/litex_boards/targets/sipeed_tang_nano.py
@@ -62,8 +62,8 @@ class _CRG(LiteXModule):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCMini):
-    def __init__(self, sys_clk_freq=48e6, with_led_chaser=True, **kwargs):
-        platform = sipeed_tang_nano.Platform()
+    def __init__(self, toolchain="gowin", sys_clk_freq=48e6, with_led_chaser=True, **kwargs):
+        platform = sipeed_tang_nano.Platform(toolchain=toolchain)
 
         # CRG --------------------------------------------------------------------------------------
         self.crg = _CRG(platform, sys_clk_freq)
@@ -90,6 +90,7 @@ def main():
     args = parser.parse_args()
 
     soc = BaseSoC(
+        toolchain    = args.toolchain,
         sys_clk_freq = args.sys_clk_freq,
         **parser.soc_argdict
     )
--- a/litex_boards/targets/sipeed_tang_nano_20k.py
+++ b/litex_boards/targets/sipeed_tang_nano_20k.py
@@ -53,13 +53,13 @@ class _CRG(LiteXModule):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCCore):
-    def __init__(self, sys_clk_freq=48e6,
-        with_led_chaser     = True,
-        with_rgb_led        = False,
-        with_buttons        = True,
+    def __init__(self, toolchain="gowin", sys_clk_freq=48e6,
+        with_led_chaser = True,
+        with_rgb_led    = False,
+        with_buttons    = True,
         **kwargs):
 
-        platform = sipeed_tang_nano_20k.Platform(toolchain="gowin")
+        platform = sipeed_tang_nano_20k.Platform(toolchain=toolchain)
 
         # CRG --------------------------------------------------------------------------------------
         self.crg = _CRG(platform, sys_clk_freq)
@@ -131,7 +131,8 @@ def main():
     args = parser.parse_args()
 
     soc = BaseSoC(
-        sys_clk_freq        = args.sys_clk_freq,
+        toolchain    = args.toolchain,
+        sys_clk_freq = args.sys_clk_freq,
         **parser.soc_argdict
     )
     if args.with_spi_sdcard:
--- a/litex_boards/targets/sipeed_tang_primer_20k.py
+++ b/litex_boards/targets/sipeed_tang_primer_20k.py
@@ -31,13 +31,14 @@ from litedram.phy import GW2DDRPHY
 # CRG ----------------------------------------------------------------------------------------------
 
 class _CRG(LiteXModule):
-    def __init__(self, platform, sys_clk_freq, with_video_pll=False):
+    def __init__(self, platform, sys_clk_freq, with_video_pll=False, with_dram=False):
         self.rst        = Signal()
         self.cd_sys     = ClockDomain()
         self.cd_por     = ClockDomain()
-        self.cd_init    = ClockDomain()
-        self.cd_sys2x   = ClockDomain()
-        self.cd_sys2x_i = ClockDomain()
+        if with_dram:
+            self.cd_init    = ClockDomain()
+            self.cd_sys2x   = ClockDomain()
+            self.cd_sys2x_i = ClockDomain()
 
         # # #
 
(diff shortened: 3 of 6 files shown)


litex_boards/platforms/sipeed_tang_nano_9k.py
```diff
@@ -99,7 +99,7 @@ class Platform(GowinPlatform):
     default_clk_period = 1e9/27e6
 
     def __init__(self, toolchain="gowin"):
-        GowinPlatform.__init__(self, "GW1NR-LV9QN88PC6/I5", _io, _connectors, toolchain=toolchain, devicename="GW1NR-9C")
+        GowinPlatform.__init__(self, "GW1NR-LV9QN88PC6/I5", _io, _connectors, toolchain=toolchain, devicename="GW1N-9C")
```
pepijndevos, line 102: This fixes the `--family` passed to Nextpnr, but might break the Gowin toolchain. Unclear what the correct option is here. Gowin is notoriously ambiguous and inconsistent with their part names.

@yrabbit what do you think about just making a symlink in nextpnr for the R devices? This is a problem people keep running into because it's just extremely confusing that the `C` is nowhere in the part number and the `R` isn't part of the database name.

litex_boards/platforms/sipeed_tang_nano_4k.py
```diff
@@ -88,10 +88,10 @@ class Platform(GowinPlatform):
 
     def __init__(self, toolchain="gowin"):
         # Educational GoWin IDE v. 1.9.8.03
-        GowinPlatform.__init__(self, "GW1NSR-LV4CQN48PC6/I5", _io, _connectors, toolchain=toolchain, devicename="GW1NSR-4C")
+        GowinPlatform.__init__(self, "GW1NSR-LV4CQN48PC6/I5", _io, _connectors, toolchain=toolchain, devicename="GW1NS-4")
         # Licensed GoWin IDE v. 1.9.7.06 Beta
         # GowinPlatform.__init__(self, "GW1NSR-LV4CQN48PC7/I6", _io, _connectors, toolchain=toolchain, devicename="GW1NSR-4C")
-        self.toolchain.options["use_mode_as_gpio"] = 1
+        #self.toolchain.options["use_mode_as_gpio"] = 1
```
pepijndevos, line 94: Somehow we don't have this setting in Apicula, or maybe it's named differently, to be investigated.

litex_boards/platforms/sipeed_tang_nano_9k.py
trabucayre: What do you think to use something like https://github.com/enjoy-digital/litex/blob/master/litex/build/xilinx/f4pga.py#L56: this allows to have this change only once is the appropriate module?

trabucayre:
LGTM. But for 20k, 25k BaseSoC call must be update with the *toolchain* param.

litex_boards/platforms/sipeed_tang_nano_4k.py
trabucayre: Why not, in a first time, using a test to add this option?
- with gowin's toolchain the current behavior is left unchanged
- with apicula's toolchain this option is not added and we have to verify/fix result ;-)

litex_boards/platforms/sipeed_tang_nano_9k.py
pepijndevos: That would be one solution but we get occasional bug reports from people who pass a wrong family or whatever and it doesn't work and Gowin part numbers are really messed up so having some safety net in Apicula / Nextpnr is preferrable I think.

Basically, there is a regex that parses the long part number, and a family argument. You'd think the long one is more specific but no, the C can only be seen in the date code. So sometimes you need both or to just specify the family. But the mapping from part to family is confusing. So that's why I'm suggesting some fallbacks for providing too much information in the family. For example R is not a separate family.

litex_boards/platforms/sipeed_tang_nano_4k.py
pepijndevos: Possible but first I'd like to look into why we don't have this option. Either we have it under a different name, or we're missing something.

pepijndevos:
- [ ] make 4k run with nolutram
- [ ] add magic sdram pins to nextpnr
- [ ] add family aliases to nextpnr
- [ ] fix toolchain in remaining gowin boards

litex_boards/platforms/sipeed_tang_nano_4k.py
pepijndevos: It's just a software check we don't have so we can make it conditional on the toolchain

litex_boards/platforms/sipeed_tang_nano_9k.py
pepijndevos: I guess the alternative is to filter the family in code, stripping elements that don't translate to a database

pepijndevos: I ended up just doing what trabucayre suggested

litex_boards/targets/sipeed_tang_nano.py
```diff
@@ -62,8 +62,12 @@ def __init__(self, platform, sys_clk_freq):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCMini):
-    def __init__(self, sys_clk_freq=48e6, with_led_chaser=True, **kwargs):
-        platform = sipeed_tang_nano.Platform()
+    def __init__(self,
+                 sys_clk_freq=48e6,
+                 with_led_chaser=True,
+                 toolchain="gowin",
+                 **kwargs):
```
trabucayre, line 69: Only one 4xspace please (has done for all method signature).

pepijndevos:
Alright so the current status is:

- Tang Nano: I think hits some regression `ERROR: No bel named X0Y5/IOBF` which I don't think is a LiteX problem
- Tang Nano 4K: works perfectly
- Tang Nano 9K: works without SDRAM
- Tang Nano 20K: builds a bitstream but don't have the board
- Tang Primer 20K: current PLL setup will require https://github.com/YosysHQ/apicula/pull/261 but works when using the Nano setup without CLKDIV.
- Tang Primer 25K: GW5A is not yet supported in Apicula so is purely forward compatibility

litex_boards/targets/sipeed_tang_nano_20k.py
```diff
@@ -71,7 +72,7 @@ def __init__(self, sys_clk_freq=48e6,
         # TODO: XTX SPI Flash
 
         # SDR SDRAM --------------------------------------------------------------------------------
-        if not self.integrated_main_ram_size:
+        if with_sdram and not self.integrated_main_ram_size:
```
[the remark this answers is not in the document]
pepijndevos: ok will undo that. I think I may have tried `--integrated_main_ram_size`  and it yelled at me so I just added a flag.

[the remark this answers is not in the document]
pepijndevos: Tang Primer 25K has such an option already but I guess because you need to specify the model?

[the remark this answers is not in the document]
trabucayre: Ah ok! I see why you have added this option. Tang 25k has no SDRAM, this option is to enable a SDRAM (sipeed or mister) connected to the header -> by default no SDRAM, if the user has plugged a module it may select the correct one

pepijndevos:
I didn't break CI right?

Any commit guidelines I should follow?

I think this is pretty much done on the LiteX side, and I'll be working on the Apicula/Nextpnr side to improve things there so we can hopefully use SDRAM.

litex_boards/platforms/sipeed_tang_nano_4k.py
```diff
@@ -91,7 +91,15 @@ def __init__(self, toolchain="gowin"):
...
-        self.toolchain.options["use_mode_as_gpio"] = 1
+
+        if toolchain=="gowin":
+            # this is just a software check that doesn't translate to hardware
+            # apicula does not implement this check, so beware of bricking your board
+            self.toolchain.options["use_mode_as_gpio"] = 1
+        else:
+            # yosys doesn't know that 4K doesn't have lutram so we tell it
+            self.toolchain._synth_opts += " -nolutram"
+
```
trabucayre, line 102: A last question here: maybe is it possible to move this part to apicula toolchain? This avoid to have to copy/paste these lines for all boards with this trick.

litex_boards/targets/sipeed_tang_primer_20k.py
```diff
@@ -58,20 +58,23 @@ def __init__(self, platform, sys_clk_freq, with_video_pll=False):
...
+                    i_CE     = self.stop,
+                    o_CLKOUT = self.cd_sys2x.clk),
+                Instance("CLKDIV",
+                    p_DIV_MODE = "2",
+                    i_CALIB    = 0,
+                    i_HCLKIN   = self.cd_sys2x.clk,
+                    i_RESETN   = ~self.reset,
+                    o_CLKOUT   = self.cd_sys.clk),
+                AsyncResetSynchronizer(self.cd_sys, ~pll.locked | self.rst | self.reset),
+            ]
```
trabucayre, line 75: This piece of code is required to use external DDR. Is the build end successfully?
Also when `with_clk` reset signal isn't driven by `self.rst` nor by `self.reset`

trabucayre:
No this PR doesn't break anything. CI fails due to a recent litedram update.

litex_boards/platforms/sipeed_tang_nano_4k.py
pepijndevos: It'd have to be spread across several tools I guess.

Yosys is currently oblivious to the specific device it's synthesizing for, so you could add a `-family` flag there or something but that doesn't really have any advantage over `-nolutram` until there are more significant differences.

We could add a `--mode-as-gpio` flag to the packer that implements the check, but like the comment said it doesn't translate to anything in hardware. It's more of a "are you sure you want to do this" thing. Given that you wrote the constraints file, I would assume the answer is "yes". All of the other `as-gpio` flags actually change bits to switch the functionality.

litex_boards/targets/sipeed_tang_primer_20k.py
pepijndevos: Ah, maybe it should be gated on `--with-sdram` then? I've only tested it without sdram because Apicula doesn't support the magic constraints yet.

In the clkdivless case it's exactly the same as Tang Nano 20k.

trabucayre: This clock source is used by the [litedram phy](https://github.com/enjoy-digital/litedram/blob/master/litedram/phy/gw2ddrphy.py#L79)

pepijndevos:
I see ci is fixed. All good now?

trabucayre:
Applied with some modifications. Thanks @pepijndevos

Issue #394, opened 2022-05-27
efinix_trion_t120_bga576_dev_kit ethernet not working


sammhho:
Hi @enjoy-digital,  may I ask what's the current support status of ethernet on the Trion t120f576 dev kit?
Is it supposed to be working, or what needs to be done to make it work?

Since there seems to be out-of-the-box support of ethernet here 
https://github.com/litex-hub/litex-boards/blob/48353735fa9471b4645cfb06557fcbdf7b4e3aaa/litex_boards/targets/efinix_trion_t120_bga576_dev_kit.py#L347

I tried to build the SoC and got a synthesis error about multiple drives on the eth_tx_rst reset signal.
I looked into the code and found that reset is built here
https://github.com/enjoy-digital/liteeth/blob/b0e72431234cbbf42a0f74f1a5f2bff145dfc043/liteeth/phy/trionrgmii.py#L177

But yet the  `pll.create_clkout` call here
https://github.com/enjoy-digital/liteeth/blob/b0e72431234cbbf42a0f74f1a5f2bff145dfc043/liteeth/phy/trionrgmii.py#L161
when called with the default `with_reset=True`, builds a reset for the domain `self.cd_eth_tx` from inverting the PLL locked https://github.com/enjoy-digital/litex/blob/c401bdd7f18f769cd59ddf77320dd8c17e50178c/litex/soc/cores/clock/efinix.py#L105

So by doing `pll.create_clkout` with `with_reset=False` on line 161 I got the SoC built successfully, 
and ran the [linux-on-litex-vexriscv](https://github.com/litex-hub/linux-on-litex-vexriscv) on it fine.
But then it won't fetch `boot.json` from tftp, so I booted thru serial, and within the Linux I can actually see the `eth0` device and successfully set ip/netmask etc., and yet `ping`-ing other IPs give no response.

BTW thx for supporting Efinix!
At Efinix we are also readying a RGMII ethernet adapter (via QSE sockets) for current Ti60 and upcoming Ti180 products, I guess I can help porting it to LiteX-boards if the `trionrgmii` code actually works.

enjoy-digital:
Hi @sammhho,

[...]

PR #608, opened 2024-09-03
efinix_trion_t120_bga576_dev_kit.py: fixed/rewire rx_ctl/tx_ctl (not compatible with DDIO mode), added message at build time

trabucayre:
As mentioned in https://github.com/litex-hub/litex-boards/issues/394#issuecomment-1140483866 *tx_ctl* pin can't be used in DDIO mode. With recent (2024.1) efinity release, ifacewriter, also, complains about *rx_ctl* for the same reason.

This PR changes rx_ctl/tx_ctl pinout by using two, not to hard, to solder pins (available through resistors).
An message is added to warn user about this requirement.


enjoy-digital:
Thanks @trabucayre

PR #609, opened 2024-09-03
targets/efinix_trion_t120_bga576_dev_kit.py: added argument to select between RMII PMOD (default) or onboard RGMII Phy

trabucayre:
With commit https://github.com/litex-hub/litex-boards/commit/7f26f3940f78fb804fa3295ac79bec40e8e945f6 a workaround has been applied for efinix_trion_t120_bga576_dev_kit.py target board to use onboard RGMII 0.
Currently users must modify target to switch between PMOD RMII and onboard RGMII.

This PR adds an argument to let user select the phy without having to modify target file.


enjoy-digital:
Thanks @trabucayre.

PR #614, opened 2024-10-04
Update tec0117 to work with Apicula

pepijndevos:
Similar to the recent Tang Nano changes.

Also makes CLKDIV optional if using 1:1 SDRAM rate and fixes the bios flashing.


trabucayre:
Applied. Thanks @pepijndevos

PR #616, opened 2024-10-08
efinix: fix reset

maass-hamburg:
fix reset on all efinix boards.
To reset the PLL a pulse is needed, which
has to be driven by a clock that is
not generated by the PLL.


enjoy-digital (approved):
Thanks @maass-hamburg!

PR #617, opened 2024-10-10
platforms/xilinx_zcu102.py: Add all SFP connectors

trabucayre:
Added pinout for SFP x 4 + common MGT ref clock


enjoy-digital:
Thanks @trabucayre!

PR #619, opened 2024-11-01
efinix_ti375_c529_dev_kit now support vexii ethernet

Dolu1990:
This can be enabled via the CPU vexiiriscv + --vexii-macsg
name=eth,address=0xF1000000,txIrq=40,rxIrq=41

Here is the linux driver : 
https://github.com/Dolu1990/litex-linux/tree/spinal-sgmac/drivers/net/ethernet/spinal

With 2 cores at 200 Mhz, in linux + iperf3, it should reach around 620 Mbits/s TX and 730 Mbits/s RX


enjoy-digital:
Thanks @Dolu1990!

PR #620, opened 2024-11-08
Update sipeed_tang_nano_4k.py

oleg-sin:
replace kB to KILOBYTE to prevent build error


trabucayre:
Applied. Thanks @oleg-sin

PR #621, opened 2024-11-09
Add support for S7 Mini (TE0890).

Philip-K:
Add a platform and target for the Trenz S7 Mini (aka TE0890).


enjoy-digital (approved):
Thanks @Philip-K!

PR #631, opened 2024-12-17
sipeed_tang_nano_20k: add SPI flash and HDMI support

andelf:
- SPI flash support, using compatible IC
- HDMI support
- `--with-rgb-led` is fixed in <https://github.com/enjoy-digital/litex/pull/2142>

SPI flash log:

```text
Initializing w25q64 SPI Flash @0x00800000...
First SPI Flash block erased, unable to perform freq test.
Memspeed at 0x800000 (Sequential, 4.0KiB)...
   Read speed: 1.1MiB/s
Memspeed at 0x800000 (Random, 4.0KiB)...
   Read speed: 674.3KiB/s
```

### DIFF
--- a/litex_boards/platforms/sipeed_tang_nano_20k.py
+++ b/litex_boards/platforms/sipeed_tang_nano_20k.py
@@ -48,6 +48,23 @@ _io = [
         IOStandard("LVCMOS33"),
     ),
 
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Great, thanks @andelf!

PR #633, opened 2024-12-26
Add hdmi conector to tang nano 20k and connector pinout.

uglyoldbob:
This pull request adds the pinout for some of the hdmi connector, and the two rows of pins on the board.


enjoy-digital:
Thanks @uglyoldbob!

PR #635, opened 2025-01-06
Add in colorlight v8.2 board support

### LINKED ISSUE #574: 5a-75b revision 8.2
hello,
i have tried to build multiple projects using liteeth on a 5a-75b 8.2 board.(liteeth, led cube from lucy and others.  unfortunatly i do not own a scope. and i have trouble getting the ethernet to work, i have found some solutions/problems and i have tried all of them. 
even tried to make one with an risc v core(litex).

has anybody got an working example with ethernet for the 5a-75b V8.2? or even just the tx/rx delay timings? i have been busy for over 4 weeks now, with no succes. is it even possible?  if i only have 100megs it will be sufficient for my usecase. 

i want to build a led screen that is capable of displaying DDP data directly from xlights,  any suggestions?

kscz:
Based on https://github.com/q3k/chubby75/blob/master/5a-75b/hardware_V8.2.md it appears that the pinout v8.0 is the same as the v8.2, the only difference is the FPGA - a Lattice ECP5 LFE5U-25F-7BG256I

The SDRAM for the 8.0 and 8.2 is the same, but the CLS was not being correctly set, this commit also corrects this.

Tested on a v8.2 board using https://github.com/enjoy-digital/colorlite

Fixes #574

### DIFF
--- a/litex_boards/platforms/colorlight_5a_75b.py
+++ b/litex_boards/platforms/colorlight_5a_75b.py
@@ -312,11 +312,11 @@ class Platform(LatticeECP5Platform):
     default_clk_period = 1e9/25e6
 
     def __init__(self, revision="7.0", toolchain="trellis"):
-        assert revision in ["6.1", "7.0", "8.0"]
+        assert revision in ["6.1", "7.0", "8.0", "8.2"]
         self.revision = revision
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Nice, thanks @kscz!

PR #636, opened 2025-01-13
Efinix Ti375 C529 Dev-Kit: Add FMC LPC connector

m-byte:
This adds the pinout of the FMC LPC connector on the Efinix Ti375 C529 Dev-Kit.


enjoy-digital:
Great @m-byte, thanks a lot!

PR #640, opened 2025-01-24
targets: efinix_ti375_c529_dev_kit: simpify and improve ddr part

maass-hamburg:
- be able to use ddr memory on all cpus, including those without a seperate memory bus.
- simplify ddr part by using ``axi_bus.get_ios()`` and ``axi_bus.connect_to_pads``.
- use the efinix python api for the DRAM Block.


enjoy-digital:
Great, thanks @maass-hamburg!

PR #642, opened 2025-01-29
Efinix Ti375 C529 Dev-Kit: Fix FMC LPC pinout

m-byte:
This fixes the pinout from the previous PR (#636), as that one accidentally used the FMC pin names, rather than the FPGA ones.


enjoy-digital:
Thanks @m-byte!

PR #645, opened 2025-02-16
Changed the programmer to fujprog on ULX3S

tarik-hamedovic:
This pull request is connected with the[ pull request](https://github.com/enjoy-digital/litex/pull/2183) I made on Litex regarding the use of `fujprog` instead of `ujprog` for progamming the ULX3S board.

Here I just changed the `radiona_ulx3s.py` file to import  and use `fujprog` instead of `ujprog`.

### DIFF
--- a/litex_boards/platforms/radiona_ulx3s.py
+++ b/litex_boards/platforms/radiona_ulx3s.py
@@ -1,217 +1,193 @@
+#!/usr/bin/env python3
+
 #
 # This file is part of LiteX-Boards.
 #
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @tarik-hamedovic, there is however an issue with the commit (it seems you replace that platform with target file). Can you fix this?

tarik-hamedovic:
My bad, I accidentally pasted the wrong file when making the fork. Now it should be okay.

enjoy-digital:
Thanks for the update, this is merged.

PR #649, opened 2025-02-24
targets/lattice_certuspro_nx_versa.py: added basic target: Leds and PCIe

trabucayre:
Only PCIe is tested

Depends on https://github.com/enjoy-digital/litepcie/pull/145


enjoy-digital:
Thanks @trabucayre!

PR #650, opened 2025-03-05
targets/litex_acorn_baseboard_mini.py: removed obsolete Platform override. Added parameter to select programmer tool

trabucayre:
The *litex_acorn_baseboard_mini* target provides a custom `Platform` inheriting from *platforms/sqrl_acorn.py* to re-implement the `create_programmer` method. But, unlike *sqrl_acorn* this variant only suppport CLE-215(+), CLE-101 is not supported.

This PR:
- removes the custom Platform to directly uses `sqrl_acorn/Platform`
- `sqrl_acorn.py` platform supports both `OpenOCD` and `OpenFPGALoader` as programmer tool. A new argument is added to allows user selecting the tool (`OpenOCD` is the default value to keep original behavior)


enjoy-digital:

PR #652, opened 2025-03-05
Add support Alinx7020 board

itisme:
1.Add support for alinx7020 board
2.implement access to the PS DDR
3.modify clock of Alinx7010 to 50 MHz according to the Alinx7010 user manual.

### DIFF
--- a/litex_boards/platforms/alinx_ax7020.py
+++ b/litex_boards/platforms/alinx_ax7020.py
@@ -73,7 +73,7 @@ ps7_config_variants = {
         "PCW_PRESET_BANK1_VOLTAGE"      : "LVCMOS 1.8V",
         #"PCW_USE_M_AXI_GP0"             : "1",
         "PCW_USE_S_AXI_HP0"             : "1",
-        #"PCW_USB0_PERIPHERAL_ENABLE"    : "1",
+        "PCW_USB0_PERIPHERAL_ENABLE"    : "1",
         #"PCW_USB0_USB0_IO"              : "MIO 28 .. 39",
         #"PCW_USB0_RESET_ENABLE"         : "1",
         #"PCW_USB0_RESET_IO"             : "MIO 46",
--- a/litex_boards/platforms/alinx_ax7020.py
+++ b/litex_boards/platforms/alinx_ax7020.py
@@ -1,9 +1,7 @@
 #
 # This file is part of LiteX-Boards.
 #
-# Copyright (c) 2022 Yonggang Liu <ggang.liu@gmail.com>,
-# Copyright (c) 2025 Neil Liu <wetone.liu@gmail.com>,
-#   change clk100 to clk50
+# Copyright (c) 2025 Neil Liu <wetone.liu@gmail.com>
 # SPDX-License-Identifier: BSD-2-Clause
 
 from litex.build.generic_platform import *
@@ -61,6 +59,29 @@ _ps7_io = [
     ),
 ]
 
+
+ps7_config_variants = {
+    "common" : {
+        "PCW_FPGA_FCLK0_ENABLE"         : "1",
+        "PCW_FPGA0_PERIPHERAL_FREQMHZ"  : "100",
+        "PCW_S_AXI_HP0_FREQMHZ"         : "100",
+        "PCW_S_AXI_HP0_DATA_WIDTH"      : "32",
+        #"PCW_UART1_BAUD_RATE"           : "115200",
+        #"PCW_EN_UART1"                  : "1",
+        #"PCW_UART1_PERIPHERAL_ENABLE"   : "1",
+        #"PCW_UART1_UART1_IO"            : "MIO 48 .. 49",
+        "PCW_PRESET_BANK1_VOLTAGE"      : "LVCMOS 1.8V",
+        #"PCW_USE_M_AXI_GP0"             : "1",
+        "PCW_USE_S_AXI_HP0"             : "1",
+        #"PCW_USB0_PERIPHERAL_ENABLE"    : "1",
+        #"PCW_USB0_USB0_IO"              : "MIO 28 .. 39",
+        #"PCW_USB0_RESET_ENABLE"         : "1",
+        #"PCW_USB0_RESET_IO"             : "MIO 46",
+        #"PCW_EN_USB0"                   : "1",
+        "PCW_UIPARAM_DDR_PARTNO"        : "MT41K256M16 RE-125",
+    },
+}
+
 # Connectors ---------------------------------------------------------------------------------------
 
 _connectors = [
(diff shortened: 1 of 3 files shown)


litex_boards/platforms/alinx_ax7020.py
```diff
@@ -0,0 +1,112 @@
...
+        #"PCW_UART1_PERIPHERAL_ENABLE"   : "1",
+        #"PCW_UART1_UART1_IO"            : "MIO 48 .. 49",
+        "PCW_PRESET_BANK1_VOLTAGE"      : "LVCMOS 1.8V",
+        #"PCW_USE_M_AXI_GP0"             : "1",
+        "PCW_USE_S_AXI_HP0"             : "1",
+        "PCW_USB0_PERIPHERAL_ENABLE"    : "1",
+        #"PCW_USB0_USB0_IO"              : "MIO 28 .. 39",
+        #"PCW_USB0_RESET_ENABLE"         : "1",
+        #"PCW_USB0_RESET_IO"             : "MIO 46",
+        #"PCW_EN_USB0"                   : "1",
```
trabucayre, line 80: Is it required to keep comments?

litex_boards/platforms/alinx_ax7020.py
```diff
@@ -0,0 +1,112 @@
...
+
+class Platform(Xilinx7SeriesPlatform):
+    default_clk_name   = "clk50"
+    default_clk_freq   = 50e6
+    default_clk_period = 1e9/default_clk_freq
+
+    def __init__(self):
+        self.ps7_config = ps7_config_variants["common"]
+        Xilinx7SeriesPlatform.__init__(self, "xc7z020clg400-2", _io, _connectors, toolchain="vivado")
+        self.add_extension(_ps7_io)
```
trabucayre, line 104: why not including _ps7_io in _io (as done for digilen_arty_z7 or digilent_zedboard)?

litex_boards/targets/alinx_ax7020.py
```diff
@@ -0,0 +1,118 @@
...
+
+# CRG ----------------------------------------------------------------------------------------------
+
+class _CRG(LiteXModule):
+    def __init__(self, platform, sys_clk_freq):
+        self.rst    = Signal()
+        self.cd_sys = ClockDomain()
+
+        self.pll = pll = S7PLL(speedgrade=-2)
+        #self.comb += pll.reset.eq(self.rst)
```
trabucayre, line 34: Why keeping a comment?

litex_boards/targets/alinx_ax7020.py
```diff
@@ -0,0 +1,118 @@
...
+        pll.register_clkin(platform.request("clk50"), platform.default_clk_freq)
+        pll.create_clkout(self.cd_sys, sys_clk_freq)
+        platform.add_false_path_constraints(self.cd_sys.clk, pll.clkin) # Ignore sys_clk to pll.clkin path created by SoC's rst.
+        
+# BaseSoC ------------------------------------------------------------------------------------------
+
+class BaseSoC(SoCCore):
+    def __init__(self, sys_clk_freq=100e6, with_led_chaser=False, with_ps_ddr=False, hp_data_width=32, **kwargs):
+        with_buttons    = False,
```
trabucayre, line 43: Is it working?

trabucayre:
With the current status it seem the PS is required with potential conflict with `--cpu-type`.
Why not using same approach than others boards based on zynq device? PS CPU is automatically added when user select zynq7000 via `--cpu-type`

litex_boards/targets/alinx_ax7020.py
itisme: This line of statement is not mandatory, but it can be added when necessary.

litex_boards/targets/alinx_ax7020.py
itisme: It can work here, but in the method BaseSoC.add_ps_module, there is a line that calls ps.add_axi_hp_slave(clock_domain="sys", data_width=hp_data_width). This depends on modifications to the Zynq-7000 module in LiteX, specifically the addition of the data_width parameter. For reference, see my other pull request: https://github.com/enjoy-digital/litex/pull/2199.

itisme:
If you use --cpu-type zynq7000, the system uses the ARM core. I hope to run the Linux system on the Zynq-7000 development board using any CPU, including RISC-V cores like VexRiscv. However, the main issue with running a larger OS on the Zynq-7000 is that DDR memory needs to be accessed from the PS side.

litex_boards/targets/alinx_ax7020.py
trabucayre: PR 2199 is now applied.
But `with_buttons = False,` is not a contructor parameter but a simple variable with an extra ","

trabucayre:
Have you code example on how to configure (in PS side I assume) the DDR to be accessible from PL?
Thanks

itisme:
Could refer https://github.com/itisme/patch_litex_linux_for_alinx7020.

trabucayre:
Thanks now it's more clear to me now.
But could you split this PR in 3 distinct PR:
1. alinx 7010 clock modification
2. alinx 7020 basic support (without DDR3 aspect)
3. DDR3 / Remapper

This will be more easy to review/integrates modifications.

Thanks.

litex_boards/platforms/alinx_ax7020.py
```diff
@@ -0,0 +1,112 @@
...
+        #"PCW_EN_USB0"                   : "1",
+        "PCW_UIPARAM_DDR_PARTNO"        : "MT41K256M16 RE-125",
+    },
+}
+
+# Connectors ---------------------------------------------------------------------------------------
+
+_connectors = [
+    ("pmodb",    "B12 B12 C12"),
+    ("pmodhdmi", "N18 P19 V20 W20 T20 U20 N20 P20 R18 R16 Y18 Y19 V16"),
```
trabucayre, line 89: It's not a PMOD but an HDMI connector -> move it to _io

litex_boards/platforms/alinx_ax7020.py
```diff
@@ -0,0 +1,112 @@
...
+    ("pmodj10",  "W19 W18 R14 P14 Y17 Y16 W15 V15 Y14 W14 P18 N17 U15 U14 P16 P15 U17 T16 V18 V17 T15 T14 V13 U13 W13 V12 U12 T12 T10 T11 A20 B19 B20 C20"),
+    ("pmodj11",  "F17 F16 F20 F19 G20 G19 H18 J18 L20 L19 M20 M19 K18 K17 J19 K19 H20 J20 L17 L16 M18 M17 D20 D19 E19 E18 G18 G17 H17 H16 G15 H15 J14 K14"),
```
trabucayre, line 91: Not PMOD: please re-organize pins to avoid long lines

trabucayre:
I have modified and fixed basic board support: see https://github.com/trabucayre/litex-boards/commit/007954b9ba2b68fa978eeedaf73949938d4e4730

I'm interested to your feedback.

PS DDR aspect will be the next step

litex_boards/platforms/alinx_ax7020.py
itisme: You are right.

itisme:
Thank you, that's clear. I create a add_ps_ddr branch and modify based on your code, [itisme@3da61bf](https://github.com/itisme/litex-boards/commit/3da61bf02695ed913d05ec96c5e1b7099a69233a), please check again.

trabucayre:
Hi @itisme 
I have modified a bit the code to integrates PS DDR in a similar way than SRAM in others targets.
commit is [here](https://github.com/trabucayre/litex-boards/commit/3f694282888cdd24693f368bfe1849bf0e6c4fad)
I have one remark:
you have:
```python
self.bus.add_slave("main_ram", axi_hp0, SoCRegion(origin=0x4000_0000, size=0x2000_0000))
```

and
```python
self.bus.slaves["main_ram"] = remap_module
```
Second line override the first.

I think it's possible to reduce code by using some existing methods like `add_remapper`

itisme:
Checking self.integrated_main_ram_size is a common practice in the system. However, if a user does not want to use main_ram, it's worth considering whether they can disable it through a parameter, rather than having to modify the code directly.

Yes, self.bus.slaves["main_ram"] will override any previous assignment. But during the system's compilation process, if "main_ram" is not first added via add_slave("main_ram", ...), it will result in a compilation error. This is because the address seen by the CPU is different. Therefore, we first add the main_ram address to the bus, and later remap it again.

As for the add_remapper or other functions related to remapping that you mentioned, I couldn't find them in the current codebase.

trabucayre:
I have updated the commit: https://github.com/trabucayre/litex-boards/commit/38238b0b36ab34a964f6d5cb5e5b0b4be2889603

`integrated_main_ram_size` is, by default, 0. So for all boards the DRAM is used and the user has set this argument to avoid this default behaviour. Here the problem is different because using PS DDR imply some steps to configure the CPU with xsct (or similar): I have moved back to `--with-ps-ddr` to simplify this step: the user must uses this argument to access the CPU DDR.

I have also removed the first `add_slave` to simply calls `add_adapter` to have a wishbone <-> Axi adapter.

itisme:
Got it, add_adapter is good. I feel this version is OK.

trabucayre:
I have applied PS_DDR with https://github.com/litex-hub/litex-boards/commit/6a66909495c7d71830a18f4e6b4519380f97afc4
I think this PR may be closed

PR #653, opened 2025-03-05
platforms/sqrl_acorn.py: fixed pcie_x4 rst_n IOStandard: BANK35 is 3.3V (conflict with LEDs)

trabucayre:
Commit 1b22061e93b2add0132cfbcd10ca0e9977a163b2 has changed PCIe `rst_n` IOStandard: before 3.3V, now 1.5V.
But the BANK35 is 3.3V, LEDs are configured in 3.3V and Vivado fails due to a conflict between IOs.
This is true for `pcie_x4` but baseboard `pcie_x1` is correctly set to 3.3V

This PR changes `pcie_x4` definition to avoid build failure.


enjoy-digital:
Thanks.

PR #661, opened 2025-04-18
Fix SFP-1 pin on sqrl_acorn.py

Mimil25:
Ref : https://github.com/enjoy-digital/litex-acorn-baseboard/blob/master/hardware/acorn.pdf


trabucayre:
Good catch! Applied. Thanks @Mimil25 !

PR #666, opened 2025-05-17
Hyvision pcie opt01 revf fix update

trabucayre:
It's a WIP to fix / updates some peripherals:
- Fixed *DDR2* `BA` pins order
- Added `SFP`/`ethernet`/`etherbone` support in the target

Notes:
- a 156.26M oscillator is available for SFP via pin `H6` but `K7_1000BASEX` PHY only support 200MHz clock frequency

Tests:
- PCIe tested: OK
- SFP 0/1 tested: OK
- DDR2: KO (When only the first chip is used memtest result is OK)

### DIFF
--- a/litex_boards/platforms/hyvision_pcie_opt01_revf.py
+++ b/litex_boards/platforms/hyvision_pcie_opt01_revf.py
@@ -69,7 +69,7 @@ _io = [
         Subsignal("ras_n", Pins("AA9"),         IOStandard("SSTL18_II")),
         Subsignal("cas_n", Pins("AB9"),         IOStandard("SSTL18_II")),
         Subsignal("we_n",  Pins("AC9"),         IOStandard("SSTL18_II")),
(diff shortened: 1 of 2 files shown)


trabucayre:
@enjoy-digital: I have updated this PR with JTAG pinout and disabled second DRAM chip: with only the first module: the memtest pass.
May be applied as it until issue is resolved.

enjoy-digital:
Thanks a lot @trabucayre, for info it's possible to select or disable DRAM chip by using PhyPadsReducer, as done here for example: https://github.com/litex-hub/litex-boards/blob/6a66909495c7d71830a18f4e6b4519380f97afc4/litex_boards/targets/qmtech_xc7k325t.py#L91. This is useful to only test with some of the modules or case like this when one module is not yet validated.

trabucayre:
Indeed. Fixed. Only 2nd odt/cs_n are required to be manually commented out to have an happy memtest.

enjoy-digital:
Good, thanks!

PR #669, opened 2025-05-24
colognechip_gatemate_evb:  SPI flash support

mmicko:
This PR fixes pin mapping for SPI and adds flash support.

Note: requires https://github.com/enjoy-digital/litex/pull/2249


trabucayre:
Applie. Thanks @mmicko !

PR #672, opened 2025-06-01
Add USB-OHCI functionality similar to what is on the digilent_arty to the nexsys_video board

spersvold:
Using PMOD connector JB supporting a PMOD like : https://github.com/nand2mario/usb_host_pmod


enjoy-digital:
Thanks @spersvold, this is merged.

PR #675, opened 2025-06-05
marble.py, marblemini.py: remove redundant files

michael-betz:
* contents of these 2 were identical with berkeleylab_marble.py and berkeleylab_marblemini.py


enjoy-digital:
Thanks @michael-betz, this is merged.

PR #676, opened 2025-06-14
berkeleylab_marble: add missing enable pin for 20 MHz VCXO

michael-betz:


enjoy-digital:
Thanks @michael-betz, this is merged.

PR #677, opened 2025-06-20
targets: try to have ethernet coherent between targets

trabucayre:
As observed in https://github.com/litex-hub/linux-on-litex-vexriscv/pull/425 ethernet integration may vary target per target:
- some targets like `digilent_arty.py` have `eth_ip`/`remote_ip`/`eth_dynamic_ip`
- some targets have a subset of these parameters
- some targets have no parameters to configure ethernet

It's true for `BaseSoC` constructor but it's also true for `add_ethernet` and `add_etherbone`

This PR is a draft to modify constructor parameters and method call to have the same structure for all targets:
- CLI arguments may or not be updated to allows users to configure target
- for some targets only `add_ethernet` is implemented: required to also adds `add_etherbone`?

Note: Since not all targets have been updated this PR is firstly to evaluate the approach before updating remaining targets.

cc @enjoy-digital @ryanhaus

### DIFF
--- a/litex_boards/targets/alibaba_vu13p.py
+++ b/litex_boards/targets/alibaba_vu13p.py
@@ -73,7 +73,7 @@ class BaseSoC(SoCCore):
                 with_etherbone         = False,
                 ethernet_port          = "qsfp0_sfp0",
                 etherbone_port         = "qsfp0_sfp0", 
-                ethernet_ip            = "192.168.1.50",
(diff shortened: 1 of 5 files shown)


enjoy-digital:
Thanks @trabucayre and sorry for the delay for this PR. I think we can merge it to have coherent ethernet parameters between targets and could eventually revisit parameters later on all targets if we want to change / cleanup things.

PR #682, opened 2025-07-12
Fix J9 pinout. Pin 1 is on the edge of the board.

0xDRRB:
Hi. J9 pinout is wrong.
Pin 1 is on the edge of the board with an arrow pointing to it, not inside.
Also, J9 has 16 pins, not 18 on the opt01 revf.
I tested the pins one by one by configuring the uart's TX and checking the output.


trabucayre:
Merged. Thanks @0xDRRB !

PR #688, opened 2025-08-29
Initial ULX5M-GS support

mmicko:
There could be some more changes like SPI flash used (since that is still what is on V002) so making PR for others to be able to test.


litex_boards/targets/intergalaktik_ulx5m_gs.py
```diff
@@ -0,0 +1,122 @@
...
+from litedram.phy import GENSDRPHY
+
+# CRG ----------------------------------------------------------------------------------------------
+
+class _CRG(LiteXModule):
+    def __init__(self, platform, sys_clk_freq):
+        self.rst        = Signal()
+        rst_n           = Signal()
+        self.cd_sys     = ClockDomain()
+        self.cd_sys_ps  = ClockDomain()
```
trabucayre, line 34: 2 spaces instead of one after self.cd_sys_ps

litex_boards/targets/intergalaktik_ulx5m_gs.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # Clk / Rst
+        clk25 = platform.request("clk25")
+        self.rst = ~platform.request("user_btn", 0)
```
trabucayre, line 38: align '=' please

PR #692, opened 2025-09-19
platforms/berkeleylab_obsidian: fix configuration of SPI flash

michael-betz:
Make writing configuration to SPI flash work.

  * only SPI_BUSWIDTH = 1 is supported
  * add name of compatible flash proxy .bit file
  * tested on hardware, with these changes the FPGA configuration can be written to flash and Obsidian boots from it with success


enjoy-digital:
Thanks @michael-betz.

PR #693, opened 2025-09-19
Add support for Icepi Zero

cheyao:
I've added support for the Icepi Zero, my ECP5 FPGA board that will be shipped out in a few months :) (https://www.crowdsupply.com/icy-electronics/icepi-zero/)


enjoy-digital:
Very nice board,  thanks @cheyao for the PR! And congrats for your successful Crowd Supply campaign :)

PR #695, opened 2025-10-03
platforms/arrow_axe5000.py: Added MKR Standard pinout peripherals

trabucayre:
The Arrow *AXE5000* / Trenz *TEI0181-01* uses Arduino MKR Form Factor for *J1*/*J2*.
In this standard some pins have usually an alternate function (`SPI`, `I2C`, `UART`, `PWM`).
This PR adds these pins groups in `_io` list.


enjoy-digital:
Thanks @trabucayre!

PR #706, opened 2025-11-10
Add support for microphase_a7_lite board

Victor-zyy:
Hi, I am victor-zyy. I just add two configurations for microphase_a7_lite board, it is a FPGA board company from china. And the FPGA is xc7a200Tfbg484-2. The VexRiscv was tested ok on that board with DDR memory and I also tested the CFU-PlayGround on that board.
Microphase_a7_lite_board


enjoy-digital:
Thanks @Victor-zyy for the board support!

PR #725, opened 2026-02-17
tangconsole: 138k and apicula support

gatecat:
WIP: adding support for the 138k variant of tangconsole and an option to bypass the PLL, needed for apicula


enjoy-digital:
Thanks @gatecat and @trabucayre. I just tried to continue the work to be able to merge it. In case some things are not fully covered, we could use another PR.

PR #729, opened 2026-03-07
GMM-7550 module support

GMM-7550:
Add support for the GMM-7550 module with a USB 3 adapter board and a memory extension module.
See <https://www.gmm7550.dev/> for the documentation and hardware design.


enjoy-digital:
Follow-up note: the latest target cleanup replaces the local SRAM Verilog bridge with the generic AsyncSRAM core proposed in enjoy-digital/litex#2502, so this PR now depends on that LiteX PR.

enjoy-digital:
Thanks. I added a few follow-up commits on top of this PR:

- 3a4d6098 cleaned up the target integration,
- 83580fe6 added the documentation entry,
- 9203d25e updated the target copyright/header,
- 721cffab switched the async SRAM support to the shared LiteX AsyncSRAM core instead of keeping a local Verilog file,
- d1b065c7 enabled bitstream loading,
- b6c38226 aligned the async SRAM pin declarations/comments,
- f2046d25 aligned the USB adapter pin declarations/comments.

The target has been smoke-tested and the parser/style checks pass. The matching LiteX core support is already merged in enjoy-digital/litex#2502, so I plan to merge this PR.

PR #731, opened 2026-03-26
targets/lambdaconcept_ecpix5.py: added usb_acm support with explicit uart_pads

trabucayre:
Fix the recent `luna_cdc_acm` update it's possible to uses `d_p/d_n` pins for USB ACM but now ULPI PHY is supported.
The ECPIX5 board has an onboard USB3300 but resources is called `ulpi` instead of default `usb`.
This PR updates the target to fix `SocCore` `uart_pads` parameters with None when `uart_name != "usb_acm"` or pass pads explicitely


enjoy-digital:
Thanks @trabucayre!

PR #734, opened 2026-04-13
decklink_quad_hdmi_recorder: Add PCIe reset line

smunaut:
Add reset line which is needed for the card to work on the Raspberry Pi 5.

The RPi5 starts the pcie ref clock at 54 MHz ! ... And then switches to 100 MHz.
During that beginning phase the reset line is asserted but if not specified in the platform file, the PLL tries to lock, goes way out of whack because of the wrong freq and when it finally goes to 100 MHz, it doesn't recover fast enough and the RPi5 has given up already ( and shutdown refclk alltogether ).

### DIFF
--- a/litex_boards/platforms/decklink_quad_hdmi_recorder.py
+++ b/litex_boards/platforms/decklink_quad_hdmi_recorder.py
@@ -40,7 +40,7 @@ _io = [
 
     # PCIe
     ("pcie_x1", 0,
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @smunaut, that's useful to know regarding the Raspberry Pi 5 behavior! Thanks for sharing.

PR #742, opened 2026-06-08
targets/terasic_atum_a3_nano.py: added with_phy_reset = False in LiteEthAgilexPHYRGMII constructor call

trabucayre:
When netboot is automatically executed at boot time first TX ARP sequence is never received by host computer/tftp server.
The RGMII phy seems requiring long delay after reset to be ready/stable.
This commit adds `with_phy_reset = False` to avoid reseting the PHY at boot time.

This PR requires https://github.com/enjoy-digital/liteeth/pull/203


enjoy-digital:
Thanks @trabucayre, this is merged.

PR #744, opened 2026-06-23
fix USB/ULPI: H11 is USB clock input, not part of ULPI

hansfbaier:
This implements the USB interfaces according to a verified working amaranth version,
see: https://github.com/amaranth-farm/deca-usb2-audio-interface/blob/main/gateware/arrow_deca.py


enjoy-digital:
Thanks @hansfbaier!

PR #752, opened 2026-07-10
litex_boards/targets/sipeed_tang_nano_9k.py: set Ibex platform regfile option to fpga

yusefkarim:
This change pairs with the recent upstream change in LiteX itself: https://github.com/enjoy-digital/litex/pull/2510

It selects Ibex's FPGA (distributed-RAM) register file on the resource-constrained GW1NR-9 so the CPU fits. Only applies specifically to the Ibex CPU on the Sipeed Tang Nano 9K (GW1NR-9), no other chip and/or board.


enjoy-digital:
Thanks @yusefkarim, this is merged.

PR #756, opened 2026-07-20
targets,platforms/colognechip_gatemate_evb: added A2 variant via --device arg, switch toolchain to peppercorn

trabucayre:


enjoy-digital:
Thanks @trabucayre!

PR #772, opened 2026-07-28
targets: uses self.cpu.set_libxil instead of explicit/dupplicated code for all boards based on zynq7000/zynmpMP

trabucayre:
This PR depends on https://github.com/enjoy-digital/litex/pull/2548

With the LiteX's PR the libxil logic has been moved to LiteX at software, soc and build levels.
The only remaining part required because specific is the `#define` required to fill `xparameters.h`.


enjoy-digital:
Thanks @trabucayre, this cleary reduces code duplication!

PR #773, opened 2026-07-30
targets: add SATA support to ECPIX-5

enjoy-digital:
## Summary

- add optional Gen2 SATA support to the ECPIX-5 target with `--with-sata`
- instantiate the ECP5 PHY on DCU1 channel 0 and expose the LiteSATA read/write core to the BIOS
- generate the fixed 150MHz SATA reference clock from a dedicated PLL, independently of the configurable system clock
- correct the platform SATA resource to expose only the DCU differential RX/TX lanes

The default target remains unchanged when SATA is not enabled.

## Hardware validation

Tested on an ECPIX-5 85F with a Toshiba DT01ACA100 HDD, using the target's default 75MHz system clock.

Build and load:

```sh
python3 litex_boards/targets/lambdaconcept_ecpix5.py \
    --with-sata \
    --build \
    --load
```

BIOS commands:

```text
sata_init
sata_read 0
sata_read 1
```

Results:

- BIOS and SDRAM initialization completed successfully
- IDENTIFY reported `TOSHIBA DT01ACA100`, 1000GB
- `Initialize SATA: OK`
- DMA reads of LBA 0 and LBA 1 completed successfully
- returned data matched the continuous pattern previously written by the LiteSATA BIST
- BIOS reached SATA filesystem probing; mounting failed as expected since the test disk has no LiteX FAT boot volume

All routed clock domains pass timing:

- system: 95.47MHz achieved / 75.01MHz required
- SATA RX: 237.81MHz achieved / 150.01MHz required
- SATA TX: 251.51MHz achieved / 150.01MHz required

The target also elaborates successfully without `--with-sata`.

## Dependency

This uses the ECP5 SATA PHY support from enjoy-digital/litesata#38. Its LiteICLink dependency, enjoy-digital/liteiclink#33, is already merged.

### DIFF
--- a/docs/boards_inventory.md
+++ b/docs/boards_inventory.md
@@ -97,7 +97,7 @@ python3 .github/scripts/generate_board_inventory.py --write
 | `kosagi_fomu` | kosagi_fomu_pvt | icestorm | `12000000.0` | - | included |
 | `kosagi_netv2` | kosagi_netv2 | vivado | `100000000.0` | Ethernet, SDCard, SPI SDCard, PCIe | included |
 | `krtkl_snickerdoodle` | krtkl_snickerdoodle | vivado | `100000000.0` | - | included |
-| `lambdaconcept_ecpix5` | lambdaconcept_ecpix5 | trellis | `75000000.0` | Ethernet, Etherbone, SDCard, Video Terminal, Video Framebuffer | included |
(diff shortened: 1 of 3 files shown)


enjoy-digital:
@gatecat: Might be of interest for you :) (We discussed that a few years ago...)

PR #401, opened 2022-06-17
Move from deprecated Symbiflow to F4PGA

robertszczepanski:
Related to [#1334](https://github.com/enjoy-digital/litex/pull/1334).

PR #774, opened 2026-08-01
targets/radiona_ulx4m_ld_v2: Disable DDR3 DM

trabucayre:
ULX4M-LD routes its DDR3 DM pins outside the DQS groups, so the ECP5 DDR output primitive cannot drive them. This uses with_dm=False from https://github.com/enjoy-digital/litedram/pull/401 to hold DM low. Since masked writes are unavailable, the target requires a non-zero L2 cache so memory writes remain full-width. Depends on enjoy-digital/litedram#401.

### DIFF
--- a/litex_boards/targets/radiona_ulx4m_ld_v2.py
+++ b/litex_boards/targets/radiona_ulx4m_ld_v2.py
@@ -147,6 +147,9 @@ class BaseSoC(SoCCore):
                 "MT41K512M16": MT41K512M16,
             }
             sdram_module = available_sdram_modules.get(sdram_device)
+            l2_cache_size = kwargs.get("l2_size", 8192)
+            if not l2_cache_size:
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Updated this PR for the finalized LiteDRAM #401 API. Commit f8915d31 replaces dm_skip with with_dm=False and fixes the constructor alignment; commit 624d8115 rejects l2_size=0 because disabling DM requires full-width writes. Validation against LiteDRAM #401: the radiona_ulx4m_ld_v2 no-compile target build passes, the zero-cache configuration fails with the intended diagnostic, and all 14 fast lint/parser tests pass. The boards CI is running; its build remains dependent on #401 being available from LiteDRAM master.

enjoy-digital:
Thanks @trabucayre!

PR #777, opened 2026-08-16
platforms/targets: add ALINX AX7035B support

rvodden:
The AX7035B is an Artix-7 XC7A35T-2FGG484I board with 256 MB of DDR3, Gigabit Ethernet, an SD card slot, QSPI flash, HDMI output driven directly from the FPGA, a 6-digit 7-segment display, an LM75 temperature sensor, a 24LC04 EEPROM and four user keys. The platform also covers the earlier AX7035, which differs only in the Ethernet PHY fitted.

Peripherals worth a note:

* HDMI is direct FPGA TMDS with no transmitter chip, so it uses VideoS7HDMIPHY and a dedicated video MMCM. Connector J6's out_en gates the 5V supply to the sink and, through a level translator pulled up to that switched rail, the DDC/EDID bus as well, so it is tied high.

* The LM75 bus has no pull-ups fitted on the board (the EEPROM bus does), so the platform requests PULLUP on those two pins. Without it SDA recovers only on pin leakage and every address appears to ACK.

* The I2C masters are the bit-banged ones rather than add_i2c_master(), because that CSR layout is the one Linux's i2c-litex driver expects.

* The 7-segment display is multiplexed in gateware; software writes one raw segment byte per digit. Its polarity is a CSR rather than hardcoded.

* The J9/J10 expansion headers are keyed by the physical pin numbers printed on the silkscreen, which start at 3 because pins 1 and 2 are GND and +5V.

Built with Vivado 2025.2 and run under linux-on-litex-vexriscv: 1280x720@60 rgb565 framebuffer at full rate, Gigabit Ethernet, SD card, QSPI flash, both I2C buses, XADC and the user keys as input events.

### DIFF
--- a/docs/boards_inventory.md
+++ b/docs/boards_inventory.md
@@ -23,6 +23,7 @@ python3 .github/scripts/generate_board_inventory.py --write
 | `aliexpress_xc7k70t` | aliexpress_xc7k70t | vivado | `90000000.0` | Ethernet, PCIe, Video Terminal, Video Framebuffer, Video Colorbars | included |
 | `alinx_ax7010` | alinx_ax7010 | - | `100000000.0` | - | included |
 | `alinx_ax7020` | alinx_ax7020 | - | `100000000.0` | - | included |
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Nice! Thanks @rvodden.

PR #794, opened 2026-09-14
targets/tang_mega: fix Etherbone construction and burst buffering

enjoy-digital:
The Tang Mega 138K Pro already supports Ethernet, but its default 16-word Etherbone record buffer cannot accommodate the bursts issued by LiteX clients. On hardware, an initial probe and small pings succeed; a 128-word read then stalls the receive path, and even ping stops responding until the FPGA is reloaded.

Set `buffer_depth=256` on both Tang Mega 138K targets so a record can contain its base address plus a complete 255-word read or write burst, matching `RemoteClient.max_burst_length`. This changes the Etherbone buffering only; the existing Gowin RGMII PHY is used.

Also correct the non-Pro target's `data_with` argument to `data_width=32`. Previously, `--with-etherbone` failed during construction with `TypeError: LiteXSoC.add_etherbone() got an unexpected keyword argument 'data_with'`.

Validation:

- Receiver simulation reproduces the old stall with a 128-word read and confirms completion of 128- and 255-word reads with the 256-word buffer.
- All six repository guardrail checks and 134 tests pass (`python3 -m pytest -q test --ignore=test/test_targets.py`).
- The non-Pro target generates successfully with `--with-etherbone --build --no-compile`; the non-Pro board was not connected for hardware testing.
- The Pro completes Gowin synthesis/place-and-route at the default 50 MHz system clock, with no reported setup/hold violations on the constrained clock domains.
- On the connected Pro at `192.168.1.50`: 1,000 pings with no loss (56- and 120-byte payloads), 200 register round trips, and eight 1 MiB DDR3 write/read passes using 128- and 255-word Etherbone bursts. The tested RAM contents are restored afterward.
- Normal `litex_server`/`RemoteClient` access passes a 1,024-word write/read transfer, including automatic splitting into 255-word bursts; `litex_cli --ident` returns the expected Pro identifier.

Hardware build command:

```sh
python3 -m litex_boards.targets.sipeed_tang_mega_138k_pro \
    --with-etherbone --eth-ip=192.168.1.50 --build --load
```


enjoy-digital:
Closing at the maintainer's request to retain LiteX-Boards' default Etherbone configuration. The connected Tang Mega 138K Pro's existing RGMII support was verified; the larger-buffer change is not being pursued. SFP-0 / 1000BASE-X support will be handled in separate PRs.

PR #858, opened 2026-09-21
Add RJ45 Ethernet for KCU105

shenki:
Support the new US_LVDS_1000BASEX PHY on the KCU105.

The device has existing Ethernet support using the SFP device. This adds support for using the existing RJ45.


enjoy-digital:
Thanks @shenki!

PR #859, opened 2026-09-23
targets,platforms: added digilent eclypse z7 board

trabucayre:
The [Digilent Eclypse Z7](https://digilent.com/reference/programmable-logic/eclypse-z7/start) is a board based on a zynq7000 (XC7Z020-1CLG484C) With 2 Pmods and 2 ZMOD


enjoy-digital:
Thanks @trabucayre!

